PR #1, opened 2019-12-13
CI

scopatz:
using conda smithy ci-skeleton


asmeurer:
I have Travis here already (this PR removes it).

scopatz:
Yep, this removes the need to manage CI configuration ourselves

asmeurer:
I don't even understand what this is doing. It seems like this is setting this repo up like a conda-forge feedstock. Where are the actual tests run? I would also be surprised if this even works since I don't have a setup.py yet.

PR #2, opened 2020-02-21
Added notebook showing # versions vs. file size for fixed array size

melissawm:
More coming soon.


asmeurer:
Should we check in the plots with the notebooks? @scopatz what do you think?

scopatz:
It is nice to have the plots when they are "done", so that it is easy for people to see the results. But there will be churn during development, so that is less great.  I think for the momement having no plots is probably better

scopatz:
@melissawm - please remove spaces from file names :wink: Thanks!

melissawm:
Done. Sorry!

melissawm:
I'll close this PR and open a new one since the notebook has changed significantly since it was first submitted.

PR #11, opened 2020-03-10
Add a test for larger-than-memory data

### LINKED ISSUE #7: Add a test for a larger than memory dataset
I'm not sure what the best way to create such a dataset in the first place is. If I just create an empty dataset it doesn't actually take up much memory in the file, but perhaps this is fine for the test.

asmeurer:
The test takes too long to run because it has to hash all the chunks of the
data. Also I'm not sure if only testing an empty dataset is sufficient.

Fixes #7.


asmeurer:
A test like this was added in #144.

PR #12, opened 2020-03-11
Python 2 support

### LINKED ISSUE #10: Python 2 support
We need to support Python 2 for now.

asmeurer:
Fixes #10.


scopatz:
This LGTM!

quarl:
FTR, since py2 support is not cheap, the plan is that for py2 we'll reimplement the latest-version lookup (expected 5-10 lines of code) outside of versioned_hdf5, to support reading the latest version only (no reading other versions and no updating).

PR #15, opened 2020-03-18
Add setup.py

### LINKED ISSUE #13: Project needs setup.py
Or other install mechanism and documentation on how to install

asmeurer:
There are no releases yet. 

Fixes #13.


setup.py
```diff
@@ -0,0 +1,29 @@
...
+    author="Quansight",
+    description="Versioned HDF5",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/Quansight/versioned-hdf5",
+    packages=['versioned_hdf5', 'versioned_hdf5.tests'],
+    license="MIT",
+    install_requires=[
+        "h5py",
+        "numpy",
```
scopatz: Isn't numpy an implicit dependency? We don't link against it.

setup.py
```diff
@@ -0,0 +1,29 @@
...
+    name="versioned-hdf5",
+    version=versioneer.get_version(),
+    cmdclass=versioneer.get_cmdclass(),
+    author="Quansight",
+    description="Versioned HDF5",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/Quansight/versioned-hdf5",
+    packages=['versioned_hdf5', 'versioned_hdf5.tests'],
+    license="MIT",
```
scopatz: I don't know why the license file says MIT when it is clearly BSD-3-Clause (which is what we should use anyway)

asmeurer: The license is MIT. I also included the license from h5py because I copied some code from there. I can switch it to BSD if that is better.

setup.py
asmeurer: We import numpy in several of the files.

scopatz: Sure, but h5py requires numpy

setup.py
scopatz: When I read the text of the license it was BSD, it just said MIT at the top. In any event, yeah, let's switch to BSD because it is a better license

setup.py
asmeurer: I generally like to list explicitly used dependencies, even if they are already implied by other dependencies.

asmeurer:
Some things here will need to be updated once #12 is merged.

LICENSE
```diff
@@ -1,24 +1,35 @@
 MIT License
```
scopatz: Probably should remove this line ;)

asmeurer:
```suggestion
BSD License
```

scopatz:
Thanks @asmeurer

PR #19, opened 2020-03-24
Fix an issue with datasets with dtype other than float64

asmeurer:


scopatz:
Looks good to me, but again needs CI and tests

asmeurer:
It looks like I can work around the Travis issue by pushing the branches up to this repo.

asmeurer:
I guess this should have had a test. I'll add one in a new PR.

PR #23, opened 2020-03-25
Fix reading a dataset from within the context manager

asmeurer:
Hoping that pushing the branch up this repo fixes the Travis thing.

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -281,6 +281,7 @@ class InMemoryDatasetID(h5d.DatasetID):
         self._shape = size
 
     def _read_chunk(self, i, mtype=None, dxpl=None):
+        # Based on Dataset.__getitem__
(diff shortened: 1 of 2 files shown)


versioned_hdf5/api.py
```diff
@@ -320,6 +322,29 @@ def write(self, mspace, fspace, arr_obj, mtype=None, dxpl=None):
 
         return data_dict
 
+    def read(self, mspace, fspace, arr_obj, mtype=None, dxpl=None):
```
scopatz: Please add a docstring

scopatz:
Code looks good! feel free to merge once there is a docstring

versioned_hdf5/api.py
asmeurer: I think it will be simpler for me to do a pass later documenting all the internal stuff. I should also probably move a lot of these things out of api.py since they aren't actually part of the public API.

PR #30, opened 2020-03-30
Allow setting the compression when creating a dataset

asmeurer:
This also removes the chunk_size parameter from stage_version(). This was the
wrong API, because the chunk size should be per-dataset, not per-version. To
set a chunk size now, set the chunks flag when calling create_dataset(), just
as you would with normal h5py.

@melissawm note the change in the API to set the chunk size here.

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -58,12 +58,12 @@ def test_stage_version_chunk_size():
                                     3*np.ones((chunk_size,))))
 
 
-        with file.stage_version('version1', '', chunk_size=chunk_size) as group:
-            group['test_data'] = test_data
+        with file.stage_version('version1', '') as group:
+            group.create_dataset('test_data', data=test_data, chunks=(chunk_size,))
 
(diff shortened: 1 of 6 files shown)


versioned_hdf5/tests/test_api.py
```diff
@@ -91,6 +91,47 @@ def test_stage_version_chunk_size():
...
+                                    2*np.ones((DEFAULT_CHUNK_SIZE,)),
+                                    3*np.ones((DEFAULT_CHUNK_SIZE,))))
+
+
+        with file.stage_version('version1', '') as group:
+            group.create_dataset('test_data', data=test_data,
+                                 compression='gzip', compression_opts=3)
+
+        with raises(ValueError):
+            with file.stage_version('version_bad') as group:
```
scopatz: These double withs can be on the same line and save a level of indentation

scopatz: I guess these are the tests, so it doesn't really matter

scopatz:
Thanks @asmeurer!

PR #31, opened 2020-04-02
Fix an issue with resizing an array that is smaller than a chunk size

### LINKED ISSUE #29: Prior data is dropped after resize?
This
```
import h5py
import numpy as np

    vf = VersionedHDF5File(f)
        sv.create_dataset('bar', data=np.array([1, 2, 3]))

    vf = VersionedHDF5File(f)
        f_bar = sv['bar']
        f_bar.resize((5,))
        f_bar[3:] = np.array([4, 5])

    vf = VersionedHDF5File(f)
    v1 = vf['1']
    print(v1['bar'][:])
```
prints
```[0 0 0 4 5]```
instead of
```[1 2 3 4 5]```

asmeurer:
Fixes #29.


ArvidJB:
Looks good, this seems to fix our issues!

PR #33, opened 2020-04-03
Add an example for writing two dimensional dense data

ArvidJB:
(which is how we actually use this in practice).

I renamed the previous dense example by appending "_old".

Note that this is not currently supported in VersionedHdf5 so I am just
writing to a regular HDF5 file.


scopatz:
@melissawm - feel free to merge whenever you are ready

PR #36, opened 2020-04-07
Fix creation of datasets

### LINKED ISSUE #32: Support groups
Creating groups in the versioned data is broken right now.

### LINKED ISSUE #34: Setting chunk size parameter to be larger than dataset size results in ValueError
Ideally, if the chunk size is too large, we should be able to ignore it or set the chunk size automatically (as the size of the initial dataset, for example). However, as other versions might have data of different sizes, this large chunk size parameter might be relevant for some versions only.

Example:

```python
>>> filename = "test.h5"                                                                                                                                                           
>>> import h5py                                                                                                                                                                    
>>> from versioned_hdf5.api import VersionedHDF5File                                                                                                                               
>>> import numpy as np
>>> f = h5py.File(filename, 'w')                                                                                                                                                   
>>> file = VersionedHDF5File(f)                                                                                                                                                    
>>> with file.stage_version("initial_version") as group:
...     val = group.create_dataset("test", data=np.random.rand(5000), dtype=(np.dtype('float64')), chunks=(2**14,)) 
... 
  File "<stdin>", line 2, in <module>
  File "/home/melissa/projects/versioned-hdf5/versioned_hdf5/api.py", line 180, in create_dataset
    data = super().create_dataset(name, **kwds)
  File "/opt/miniconda3/envs/vhdf5/lib/python3.8/site-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/opt/miniconda3/envs/vhdf5/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 104, in make_new_dset
    raise ValueError(errmsg)
ValueError: Chunk shape must not be greater than data shape in any dimension. (16384,) is not compatible with (5000,)
>>> 
```

asmeurer:
Prior to this, datasets are created in the group of the previous version. But
this actually corrupts the previous version if any new datasets are created.
Instead, we should copy everything over from the previous version. It also
means we shouldn't actually call create_dataset() on the h5py Group object,
but rather fake it so that it goes through our machinery.

This isn't completely fixed yet. There are still some bugs, and we need to
handle groups properly. There are also some tests that need to be fixed, as
well as new tests that need to be written.

This will eventually fix #32 and fix #34.

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -1,7 +1,10 @@
-from h5py import Empty, Dataset, Group, h5d, h5i, h5s
-from h5py._hl.base import phil
+from h5py import Empty, Dataset, Datatype, Group, h5d, h5i, h5p, h5s, h5t
+from h5py._hl.base import guess_dtype, phil
+from h5py._hl.dataset import _LEGACY_GZIP_COMPRESSION_VALS
+from h5py._hl import filters
 from h5py._hl.selections import select
 from h5py._hl.vds import VDSmap
+
 
 import numpy as np
 
@@ -11,8 +14,9 @@ import datetime
 import math
 
 from .backend import initialize
-from .versions import (create_version, get_nth_previous_version,
-                       set_current_version, all_versions)
+from .versions import (create_version_group, create_version,
+                       get_nth_previous_version, set_current_version,
+                       all_versions)
 from .slicetools import s2t, slice_size, split_slice
 
 class VersionedHDF5File:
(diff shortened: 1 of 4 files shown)


scopatz:
CI is failing, but the code looks good to me

versioned_hdf5/api.py
```diff
@@ -206,6 +216,118 @@ def _get(name, item):
...
+
+    # Validate chunk shape
+    if isinstance(chunks, int) and not isinstance(chunks, bool):
+        chunks = (chunks,)
+    # if isinstance(chunks, tuple) and any(
+    #     chunk > dim for dim, chunk in zip(tmp_shape, chunks) if dim is not None
+    # ):
+    #     errmsg = "Chunk shape must not be greater than data shape in any dimension. "\
+    #              "{} is not compatible with {}".format(chunks, shape)
+    #     raise ValueError(errmsg)
```
scopatz, line 270: Please remove dead code or provide context

asmeurer:
This doesn't fix groups yet, but I am going to merge this and fix that in another PR. This already fixes #34 and I suspect many other issues that have not been spotted yet.

PR #37, opened 2020-04-07
Fix resize computation when last chunk does not align with chunk_size

ArvidJB:
Before this change the current resize logic is incorrectly resizing the
array in data_dict by concatenating with an all zero array of the full
length instead of just the remainder in the last chunk.

This fixes https://github.com/Quansight/versioned-hdf5/issues/35


scopatz:
LGTM. CC @asmeurer

PR #38, opened 2020-04-07
Refactor

asmeurer:
This PR just moves things around.

Move non-public things out of api.py

Move h5py wrapper classes into a new wrappers.py and spaceid_to_slice() into
slicetools.py.

Move the setup() test function into a helpers file.


scopatz:
Thanks!

PR #39, opened 2020-04-08
Don't require the version data to be at the root of the file

asmeurer:


scopatz:
LGTM - can you also add a test for this please?

PR #43, opened 2020-04-10
Add codecov to Travis

asmeurer:


scopatz:
Do you actually find codecov useful?

asmeurer:
It's not great, but I'm hoping it can help. It's easier than running the coverage report locally. I do find the reports on PRs to be mostly annoying and I will probably turn those off. 

I also guess this doesn't actually work yet. I don't know what else I need to do https://codecov.io/gh/Quansight/versioned-hdf5

scopatz:
I am fine merging this as-is then.

PR #44, opened 2020-04-10
Add InMemoryArrayDataset

asmeurer:
This is used when the dataset is created from an array. That way, we don't
have to create a fake Dataset on the file so that we can use InMemoryDataset.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -121,7 +121,7 @@ def write_dataset_chunks(f, name, data_dict):
         ds[t2s(raw_slice)] = data_s
     return slices
 
-def create_virtual_dataset(f, version_name, name, slices):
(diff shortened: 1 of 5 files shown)


asmeurer:
This still needs more tests.

scopatz:
Implementation looks fine, at this point

asmeurer:
This now adds support for attrs. For now they are just represented in the API as a dictionary. If we need to, we can make a wrapper so they act more like the h5py AttributeManager object. One effect of this is that the type checking on attrs won't happen until the version is committed (this sort of thing, where errors only happen when the version is committed, happens in some other places as well).

ArvidJB:
Looks good, I just tried those changes and they seem to fix the issues I had!
👍

scopatz:
Thank @asmeurer!

PR #46, opened 2020-04-22
Test to keep API compatible with generate_data.py

melissawm:
Please review it to make sure this makes sense.

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -268,6 +268,60 @@ def test_create_dataset():
         assert list(f['/_version_data/versions/version1']) == list(file['version1']) == ['test_data']
         assert list(f['/_version_data/versions/version2']) == list(file['version2']) == ['test_data', 'test_data2']
 
+
+def test_changes_dataset():
+    # Testcase similar to those on generate_data.py
+    test_data = np.concatenate((np.ones((2*DEFAULT_CHUNK_SIZE,)),
+                                2*np.ones((DEFAULT_CHUNK_SIZE,)),
+                                3*np.ones((DEFAULT_CHUNK_SIZE,))))
+    name = "testname"
+    
+    with setup() as f:
+        file = VersionedHDF5File(f)
+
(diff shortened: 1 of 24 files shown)


asmeurer:
This looks fine. I think there is some duplication here from https://github.com/Quansight/versioned-hdf5/pull/39.

scopatz:
There are some conflicts here now

melissawm:
Should be ok now

ArvidJB:
Having some unit tests to make sure that the test cases in generate_data.py continue to work is a great idea. I cannot currently run those tests myself...

scopatz:
Looks like CI failed

versioned_hdf5/tests/test_api.py
```diff
@@ -268,6 +268,46 @@ def test_create_dataset():
         assert list(f['/_version_data/versions/version1']) == list(file['version1']) == ['test_data']
         assert list(f['/_version_data/versions/version2']) == list(file['version2']) == ['test_data', 'test_data2']
 
+
+def test_changes_dataset():
+    # Testcase similar to those on generate_data.py
```
asmeurer, line 273: Is it possible to literally re-use generate_data, but only do 2 or 3 transactions? That way we always stay up-to-date with it.

ArvidJB: If we expose `num_transactions` as a parameter that should work. For example
```
def _test_mostly_appends_sparse(self, num_transactions):
    # ...
def test_mostly_appends_sparse_small(self):
    self._test_mostly_appends_sparse(5)
def test_mostly_appends_sparse_large(self):
    self._test_mostly_appends_sparse(2500)

```

melissawm: This is what I do in the analysis branch (exposing `num_transactions` and other parameters) to automate performance tests.

The test I've written there is a simplified version of generate_data just because I wanted to avoid having helper functions like `_modify_dss_sparse` and `modify_dss_dense` here. But the operations are basically the same.

melissawm:
This should be ready for merging after review.

PR #49, opened 2020-04-24
Use powerlaw for insert/delete distribution in generate_data tests

ArvidJB:
This change makes sure that inserts/deletes only happen on the "most
recent" data. It also changes the logic to only load and write the slice
of the Dataset which is actually impacted which should improve the
performance.

I also refactored the duplicated code between the versioned and regular
test case out into a common base class.

Note that I cannot verify that the versioned tests still succeed since
they no longer work against master, unfortunately.

### DIFF
--- a/analysis/generate_data_base.py
+++ b/analysis/generate_data_base.py
@@ -1,0 +1,319 @@
+from __future__ import (absolute_import, division, print_function, with_statement)
+
+import abc
+import datetime
+import random
+from unittest import TestCase
+
+import numpy as np
+import scipy.stats
+
+
+class TestDatasetPerformanceBase(TestCase, metaclass=abc.ABCMeta):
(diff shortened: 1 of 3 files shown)


scopatz:
@melissawm - can you please review? The code looks fine to me

scopatz:
But this is really your area :wink:

ArvidJB:
Thanks @scopatz !
@melissawm, this should not make any big changes other than changing inserts/deletes from
"insert/delete anywhere in the array and then write it back via `ds[:] = a`
to
"bias insertion points heavily toward the end, figure out the first inserted/deleted point `minr`, only update `ds[minr:] = a`"
The additional changes are just refactorings, let me know if this conflicts with any of the changes you have made to run your analysis.

PR #40, opened 2020-04-08
Groups support

asmeurer:
Fixes #32 
Fixes #48

PR #51, opened 2020-04-29
Delete the version if the context manager body raises an exception

asmeurer:
Previously it only did this if the context manager exit code raised an
exception.


melissawm:
This PR fails on `generate_data.py` but I think this is because #40 needs to be merged first, is that right?

asmeurer:
Yeah I believe #40 is a prerequisite for generate_data working.

PR #53, opened 2020-04-30
Fix unnecessarily quadratic algorithm to be linear

### LINKED ISSUE #52: test_mostly_appends_sparse is very slow
Running `test_mostly_appends_sparse` against regular HDF5 without versioning takes about 2-3 seconds. Running it with versioning takes hours. Both are running 2500 transactions.

asmeurer:
Fixes #52

CC @ArvidJB 

Before this with 100 transactions, test_mostly_appends_sparse takes 30 seconds. With this branch, it takes 2.7 seconds. I didn't test master with the full 2500 transactions but on this branch it takes 8:40 (@ArvidJB reported it took hours for him on master). I suspect there are still performance gains that can be made. 

Note that for me at least, the generate_data script is broken unless you first merge my branch from #40 into this one.

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -18,7 +18,7 @@ from collections import defaultdict
 import math
 import posixpath as pp
 
-from .slicetools import s2t, slice_size, split_slice, spaceid_to_slice
+from .slicetools import s2t, t2s, slice_size, split_slice, spaceid_to_slice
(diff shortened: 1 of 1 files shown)


ArvidJB:
Awesome, performance is much improved! Let's try to merge this and #40 into master (do you know why the tests are failing) and then I will try to play with it a bit more.

PR #57, opened 2020-05-07
Performance analysis notebooks

melissawm:
This is latest, cleaned up version of the notebooks to show performance analysis for Versioned HDF5. Including
- Compression tests
- Chunk size comparison
- Some IO tests
- `performance_tests.py`, a script to run tests in batch mode.


melissawm:
This can be merged once you approve, @asmeurer 

Thanks!

generate_data_deterministic.py
```diff
@@ -0,0 +1,554 @@
...
+                                                             num_inserts)
+                else:
+                    cls._modify_dss_sparse_deterministic(key0_ds, key1_ds, val_ds,
+                                                         num_rows_per_append,
+                                                         num_changes,
+                                                         num_deletes,
+                                                         num_inserts)
+
+                t = time.time()
+                times.append(t-told)
```
asmeurer, line 124: It might remove some noise if you call `f.flush()` after each loop. Otherwise h5py may randomly decide to flush in the middle of some operations, which would cause the time to spike for that run.

PR #59, opened 2020-05-08
Make InMemoryGroup singletonized per bind

### LINKED ISSUE #58: Error with more than one nesting level
This does not work
```
    vf = VersionedHDF5File(f)
    with vf.stage_version('v0') as sv:
        sv.create_dataset('a/b', data=np.array([1, 2, 3]))
        print(sv['a/b'][:])
        sv.create_dataset('a/c/d', data=np.array([1, 2, 3]))
        print(sv['a/b'][:])
```
because creating 'a/c/d' seems to delete group 'a', probably because this line
```        dirname, data_name = pp.split(name)```
only splits off one level.

asmeurer:
Otherwise create_group could create a new object for a group that already has
been made, losing the in-memory datasets and groups that it has on it.

This still needs tests.

Fixes #58.

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -711,6 +711,61 @@ def test_groups():
         assert_equal(version['group1']['group2/test_data'], data)
         assert_equal(version['group1/group2/test_data'], data)
 
+        with file.stage_version('version6', '') as group:
+            group.create_dataset('group1/test_data1', data=data)
+            group.create_dataset('group1/group2/test_data2', data=2*data)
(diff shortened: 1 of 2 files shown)


asmeurer:
Test added.

ArvidJB:
Looks good. Is there a way to tie the lifetime of an `InMemoryGroup` to its parent group instead of using a global map?

asmeurer:
Yeah that's what the `create_group` logic was trying to do, but it was incorrect. So I figured it would be safer to do it this way, rather than trying to do careful bookkeeping everywhere.

PR #61, opened 2020-05-13
Support multiple dimensions (part 1)

asmeurer:
Currently the dataset can only be expanded (chunked) in the first dimension. Support for
expanding in all dimensions will come later. 0-dimensional arrays are also not
yet supported.

This passes all tests for single dimension arrays, but still needs tests for
actual multidimensional arrays.

### DIFF
--- a/versioned_hdf5/slicetools.py
+++ b/versioned_hdf5/slicetools.py
@@ -25,12 +25,19 @@ def split_slice(s, chunk):
         new_step = step
         yield i, Slice(new_start, new_stop, new_step)
 
-def split_chunks(shape, chunk_size):
-    if len(shape) > 1:
-        raise NotImplementedError
+def split_chunks(shape, chunks):
+    if len(shape) != len(chunks):
+        raise ValueError("chunks shape must equal the array shape")
+    if len(shape) == 0:
+        raise NotImplementedError("Scalar datasets")
+
+    if any(i != j for i, j in zip(shape[1:], chunks[1:])):
+        raise NotImplementedError("Chunking over any dimension other than the first is not yet implemented")
+
+    chunk_size = chunks[0]
 
     for i in range(math.ceil(shape[0]/chunk_size)):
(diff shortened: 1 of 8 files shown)


ArvidJB:
I started playing with it by running `test_mostly_appends_dense` after commenting out all inserts and deletes for now. I ran into some issues:
* `create_virtual_dataset` stumbles on this line
```
    if not all(isinstance(s, Slice) for s in slices):
        raise NotImplementedError("Chunking in other than the first dimension")
```
* resize does not seem to work and fails at this line
```
                assert a.shape[0] == old_shape % chunk_size
```
in `set_extent`.

asmeurer:
When inserting rows, it tries to insert into the second dimension, but only the first is supported. I've pushed a fix with a better message for that. 

I think there is a bug though even when not doing inserts, which I am looking into.

ArvidJB:
I just tried this out with your latest changes and the dense tests now pass if you do not insert/delete from the second dimension.

versioned_hdf5/slicetools.py
```diff
@@ -25,12 +25,19 @@ def split_slice(s, chunk):
         new_step = step
         yield i, Slice(new_start, new_stop, new_step)
 
-def split_chunks(shape, chunk_size):
-    if len(shape) > 1:
-        raise NotImplementedError
+def split_chunks(shape, chunks):
```
scopatz, line 28: Please add a docstring for public functions

scopatz:
LGTM - CC @melissawm

versioned_hdf5/slicetools.py
asmeurer: Only the stuff in api.py should be considered public. Also the stuff in slicetools.py will be replaced with ndindex.

scopatz: > Also the stuff in slicetools.py will be replaced with ndindex.

This is fair.

Issue #60, opened 2020-05-13
Creating empty dataset with fillvalue fails


ArvidJB:
This code snippet
```
with h5py.File('foo.h5', 'w') as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version('0') as sv:
        sv.create_dataset('bar', shape=(6,), dtype=np.dtype('int64'), data=None,
                          maxshape=(None,), chunks=(10000,), fillvalue=0)
```
fails with
```
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
```
There seems to be some confusion where in `_make_new_dset` we use the correct dtype but then the `InMemoryDataset` has dtype 'object'?

asmeurer:
There are two issues here

- Setting a fill value
- Creating an empty dataset

Supporting fill values should be straightforward to fix. We just need to store the fill value and use it where appropriate (there are a few places in the code that implicitly assume the fill value is 0). 

Supporting empty datasets may be harder. I will have to think about how they can work with the current backend design. We can store the full array, but ideally we would like to store an empty dataset without storing the empty parts of it, just as the base hdf5 does.

asmeurer:
Perhaps for empty chunks we can store a special value in the data hashtable.

asmeurer:
Actually the virtual datasets will already do the right thing with missing entries. So maybe this is just a question of removing the assumption that all the chunks in the virtual dataset point to a raw data chunk.

PR #62, opened 2020-05-20
Add support for fill values

asmeurer:
We still don't support empty datasets, so the fill value only affects resizes.

Fixes partially #60


scopatz:
This seems Ok to merge to me. Is it safe to?

scopatz:
This has conflicts now

scopatz:
LGTM, thanks!

PR #63, opened 2020-05-21
Support for sparse datasets

asmeurer:
Sparse datasets have chunks that are not stored directly in the file. The data
in those chunks is assumed to be equal to the fill value.

This supports writing sparse datasets equal to the fill value
(create_dataset() with data=None), but it does not yet support writing partial
sparse data. Reading values is also not yet supported except when the file is
closed and reopened. Support for these features requires some features to be
added to ndindex first.

This is based on #62, which can be merged independently.

See issue #60.


versioned_hdf5/backend.py
```diff
@@ -28,16 +28,17 @@ def create_base_dataset(f, name, *, shape=None, data=None, dtype=None,
...
-                                   compression_opts=compression_opts)
+                                   compression_opts=compression_opts,
+                                   fillvalue=fillvalue)
 
     dataset.attrs['chunk_size'] = chunk_size
     return write_dataset(f, name, data, chunk_size=chunk_size)
 
 def write_dataset(f, name, data, chunk_size=None, compression=None,
-                  compression_opts=None):
+                  compression_opts=None, fillvalue=None):
```
scopatz, line 38: These methods need docstrings

versioned_hdf5/wrappers.py
```diff
@@ -310,14 +322,63 @@ def chunks(self):
...
+        self.chunk_size = chunk_size or DEFAULT_CHUNK_SIZE
+
+        # This works like a mix between InMemoryArrayDataset and
+        # InMemoryDatasetID. Explicit array data is stored in a data_dict like
+        # with InMemoryDatasetID, but unlike it, missing data (which equals
+        # the fill value) is omitted.
+        self.data_dict = {}
+
+    @classmethod
+    def from_dataset(cls, dataset):
```
scopatz, line 346: These methods also need docstrings

asmeurer:
The work here isn't finished yet, but it is also going to require some ndindex improvements to be usable for anything nontrivial.

asmeurer:
This also will have some nontrivial merge conflicts with https://github.com/Quansight/versioned-hdf5/pull/61. So we should determine which should be finished first.

ArvidJB:
I have a workaround for #60 so I would prioritize #61, especially since it (#61) seemed to mostly work when I tried it out.

scopatz:
This has conflicts now

asmeurer:
@ArvidJB what would be a typical shape and chunk size for a sparse (empty) dataset for your use-case? You mentioned that you can work-around this by simply storing them densely. Does that mean they aren't too large to actually fit in memory?

The reason I ask is that currently when a dataset is indexed, we enumerate all the chunks to see which are included in the index. However, if the total number of chunks for a densely represented dataset would be large, this may be too inefficient to do. But if the total number of chunks isn't too large (say, less than 10000), then this isn't a huge issue. It may be some more work to make this more efficient, so if it isn't actually required for your use-case, I may prefer to skip doing so for now. But if it is an issue, then it's important that I do make it more efficient, because doing something like enumerating 1000000 chunks would make every dataset operation take over a minute.

asmeurer:
c.f. https://github.com/Quansight/ndindex/issues/45

ArvidJB:
The typical usage of this is for constant datasets. The most common example is when we store a "mask" array to indicate whether any of the entries in an array is `None`. Normally that is pretty rare, most of the time there will be no `None`s and the array will be constant `False`.

Typically these would have the same shape as the "data" arrays. Normally for one-dimensional tables the typical chunk size is 10,000 and arrays are up to length 100,000,000. For two dimensional arrays the chunk size is probably between 100x100 and 500x500. I think the largest arrays here are probably 10,000x10,000.

scopatz:
The code looks good to me

asmeurer:
OK, I think the current behavior should have acceptable performance at those sizes. You'll have to let me know if you notice performance issues with indexing.

asmeurer:
This wasn't actually ready to go yet.

scopatz:
Oh whoops. Feel free to revert if you need

PR #72, opened 2020-06-11
Work on multidimensional datasets

asmeurer:
Requires ndindex master.


asmeurer:
I am going to use this PR to add full multidimensional support. I'll post an update when things are ready to test.

asmeurer:
For the initial pass here, I am going to disallow data shapes that are not multiples of the chunk other than in the first dimension. Supporting this will require changing the backend design a bit, and I need to think about how to properly do it.

asmeurer:
I've done an ndindex release so it is no longer required to use the ndindex git version here. ndindex 1.3 is required.

asmeurer:
I have basic multidimensional support working. With some caveats:

- All multidimensional datasets must specify a chunk size when creating the dataset, using `create_dataset`.
- `maxshape` is not implemented. 
- Certain types of advanced indexing are not yet supported. These need to be supported in the upstream ndindex code, so it will be useful to know what sorts of things are missing that are needed.
- Resizing multidimensional datasets is not yet implemented
- It requires https://github.com/Quansight/ndindex/pull/57 for all cases to work (particularly for integer indexing)
- Still not properly tested.

ArvidJB:
I tried this code, but I cannot create a multi-dimensional dataset:
```
    vf = VersionedHDF5File(f)
        g = sv.create_group('bla')
        g.create_dataset('bar', data=np.array([[1, 2], [2, 3]]),
                         maxshape=(None, None), chunks=(1000, 1000))
```
fails with
```NotImplementedError: chunks must be specified for multi-dimensional datasets```
It seems like the problem is that the `chunks` are taken from the "root" group?

**Comment by asmeurer** [comment]
Ah, it looks like there is a bug with subgroups. The bug also affects the compression options as well.

**Comment by asmeurer** [comment]
I've pushed up a fix.

**Comment by ArvidJB** [comment]
Thanks @asmeurer I am now getting further. The next stumbling block is the `resize` method which is not yet implemented.

**Comment by asmeurer** [comment]
OK, I now have resizing working. This is now ready for review. 

A note for the reviewers: I have an exhaustive test for multidimensional resizing. The test is a bit slow because it tests a lot of combinations, but I'd like to keep it because it caught a lot of bugs. There are two versions of the test. The version that is there now takes about 8 minutes to run and recreates the test file thousands of times throughout the test. There is an equivalent version that reuses the same file for the entire test. It only takes about 2 minutes to run, but the final test file at the end of the test is about 1 GB.

**Comment by asmeurer** [comment]
Also the tests here are probably going to fail until I get https://github.com/Quansight/ndindex/pull/57 released, which is my next priority.

**Comment by ArvidJB** [comment]
I just tested this and it seems to (mostly) work!
Two things I saw:
* https://github.com/deshaw/versioned-hdf5/issues/97 (I am skipping the fancy indexing for now)
* `resize` does not support arguments of `numpy.int64` anymore

**Comment by asmeurer** [comment]
> resize does not support arguments of numpy.int64 anymore

Can you clarify what doesn't work here? I tried passing a tuple of int64 as the argument to resize, and it seems to be working. What doesn't work is an integer (not tuple) argument, like `data.resize(10)` instead of `data.resize((10,))`. However, this also didn't work before this PR. This behavior is copied from h5py, although we could certainly make it work if that is desired.

**Comment by ArvidJB** [comment]
> > resize does not support arguments of numpy.int64 anymore
> 
> Can you clarify what doesn't work here? I tried passing a tuple of int64 as the argument to resize, and it seems to be working. What doesn't work is an integer (not tuple) argument, like `data.resize(10)` instead of `data.resize((10,))`. However, this also didn't work before this PR. This behavior is copied from h5py, although we could certainly make it work if that is desired.

This is for a proper shape tuple:
```
import h5py
import numpy as np
    vf = VersionedHDF5File(f)
        data_group = sv.create_group('data')
        data_group.create_dataset('bar', data=np.arange(4))
    vf = VersionedHDF5File(f)
        data_group = sv['data']
        data_group['bar'].resize((np.int64(10),))
 ```
results in
```     
  File "/usr/local/python/python-3.7/std/lib64/python3.7/site-packages/IPython/core/interactiveshell.py", line 3375, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-b3539d153545>", line 5, in <module>
    data_group['bar'].resize((np.int64(10),))
  File "/codemill/bessen/fork/base/src/python/versioned_hdf5/wrappers.py", line 423, in resize
    data = np.full(c.newshape(size), self.fillvalue, dtype=self.dtype)
  File "/codemill/bessen/fork/base/src/python/ndindex/tuple.py", line 329, in newshape
    newshape.extend(list(s.newshape(shape[i])))
  File "/codemill/bessen/fork/base/src/python/ndindex/slice.py", line 363, in newshape
    idx = self.reduce(shape)
  File "/codemill/bessen/fork/base/src/python/ndindex/slice.py", line 308, in reduce
    if len(shape) <= axis:
TypeError: object of type 'numpy.int64' has no len()
```

ArvidJB:
Also, there now seems to be a bug in `set_chunks` which leads to the wrong chunks being used:
```
import h5py
import numpy as np

    vf = VersionedHDF5File(f)
        data_group = sv.create_group('data')
        data_group.create_dataset('bar', data=np.arange(4))

    vf = VersionedHDF5File(f)
        data_group = sv['data']
        data_group.create_dataset('props/1/bar', data=np.arange(0, 4, 2))
```
which complains about
```
  File "/u/bessen/opensource/versioned-hdf5/set_chunks_dict_bug.py", line 15, in <module>
    data_group.create_dataset('props/1/bar', data=np.arange(0, 4, 2))
    next(self.gen)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/api.py", line 170, in stage_version
    timestamp=timestamp)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/versions.py", line 96, in commit_version
    raise NotImplementedError("Specifying chunk size with dict data")
NotImplementedError: Specifying chunk size with dict data
```

ArvidJB:
Actually, here's a pretty simple test which fails:
```
f = h5py.File('foo.h5', 'w')
vf = VersionedHDF5File(f)
    sv.create_dataset('bar', data=np.full((10, 10, 10), 5), chunks=(20, 20, 20))

with vf.stage_version(str(i)) as sv:
    new_shape = (28, 29, 11)
    sv['bar'].resize(new_shape)
```
with the following error
```
  File "/u/bessen/opensource/versioned-hdf5/resize_dense_error.py", line 16, in <module>
    sv['bar'].resize(new_shape)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/wrappers.py", line 425, in resize
    data[old_shape_idx.as_subindex(c).raw] = a
ValueError: could not broadcast input array from shape (10,0,10) into shape (0,9,11)
```

asmeurer:
Thanks. I think I may have already fixed this in the work I did today. If you pull the latest version of my ndindex branch, the error goes away.

ArvidJB:
> Thanks. I think I may have already fixed this in the work I did today. If you pull the latest version of my ndindex branch, the error goes away.

Confirmed, using https://github.com/asmeurer/ndindex/tree/as_subindex-improvements it seems to work!

asmeurer:
ndindex 1.3.1 is now released. You no longer need to use the development branch. It also fixes the issue with using numpy integers in the shape.

versioned_hdf5/tests/test_api.py
```diff
@@ -577,6 +578,64 @@ def test_resize_unaligned():
...
+def test_resize_multiple_dimensions():
+    # Test semantics against raw HDF5
+
+    shapes = range(5, 25, 5) # 5, 10, 15, 20
+    chunks = (10, 10, 10)
+    with setup() as f:
+        file = VersionedHDF5File(f)
+        for i, (oldshape, newshape) in\
+            enumerate(itertools.combinations_with_replacement(itertools.product(shapes, repeat=3), 2)):
+            print(oldshape, newshape)
```
melissawm, line 590: Is this print supposed to stay in the test?

versioned_hdf5/tests/test_api.py
```diff
@@ -1168,6 +1224,43 @@ def test_multidimensional():
         assert version2['test_data'][0, 1] == 2
         assert_equal(version2['test_data'][()], data2)
 
+        data3 = data.copy()
+        data3[0:1] = 3
+
+        with file.stage_version('version3', 'version1') as g:
```
melissawm, line 1230: Is there a special reason why you used `version1` as the previous version? Just out of curiosity

versioned_hdf5/tests/test_api.py
asmeurer: Oops, no.

asmeurer: Although it would be cool if we could parameterize this test, similar to the InMemoryArrayDataset test in test_wrappers.py. The test takes 2 minutes to run and it would be nice to see the progress of it in the test runner output. The problem is that we have to reuse the same file for all the parameters. Recreating a file for each iteration makes the test take 10 times longer (https://github.com/deshaw/versioned-hdf5/pull/72#issuecomment-658452371). I don't know if pytest fixtures support reusing a tempfile fixture across parameterizations.

versioned_hdf5/tests/test_api.py
asmeurer: If you don't specify a previous version, it defaults to the most recently created version, which in this case would be version2. However, want to test against modifications from the base data, which is in version1. Otherwise I would have to also include the modifications that I did in version2 above, which would make the test more complicated (admittedly not too much more complicated, but it's easier to see what's going wrong when things break if you don't have to recreate several layers of changes).

scopatz:
LGTM  - @asmeurer, feel free to merge when ready

scopatz:
:tada: Congrats!

PR #80, opened 2020-06-12
Update README with History and Licence

mlucool:
D. E. Shaw logo used under these terms: https://www.deshaw.com/tmug


asmeurer:
Looks OK to me. Maybe @scopatz should signoff too.

asmeurer:
Also, is this what you had in mind for the sponsorship language for the release, or should we do more there? The README is currently configured to be used as the long description on PyPI (although it doesn't appear in the documentation at https://deshaw.github.io/versioned-hdf5/).

PR #81, opened 2020-06-12
Adds close method to VersionedHDF5File objects. Updates docs.

### LINKED ISSUE #45: Add close() method to VersionedHDF5File
As mentioned in https://github.com/Quansight/versioned-hdf5/pull/42

melissawm:
Please let me know if there's a better approach.

Closes #45

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -162,17 +162,10 @@ class VersionedHDF5File:
         Make sure the VersionedHDF5File object is no longer reachable.
         """
         if not self._closed:
-            self.__del__()
-            self._closed = True
-
-    def __del__(self):
-        """
-        Prevent access to VersionedHDF5File data if the file is closed.
-        """
-        if not self._closed:
             del self.f
             del self._version_data
             del self._versions
+            self._closed = True
 
     def __repr__(self):
         """
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -15,6 +15,7 @@ from .versions import (create_version_group, commit_version,
                        all_versions, delete_version)
 from .wrappers import InMemoryGroup
 
+
 class VersionedHDF5File:
     """
     A Versioned HDF5 File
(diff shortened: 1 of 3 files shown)


versioned_hdf5/api.py
```diff
@@ -150,3 +156,32 @@ def stage_version(self, version_name: str, prev_version=None,
         except:
             delete_version(self.f, version_name, old_current)
             raise
+
+    def close(self):
+        """
+        Make sure the VersionedHDF5File object is no longer reachable.
+        """
+        if not self._closed:
+            self.__del__()
```
asmeurer, line 165: Should it call self.f.close()?

asmeurer:
What is the purpose of the `__del__` method here?

versioned_hdf5/api.py
melissawm: I was not sure if this should be done silently.

melissawm:
> What is the purpose of the `__del__` method here?

It was about not forgetting to `close` the object after finishing but I know this might not be the best solution... Also since if the `h5py` file is closed the versioning data gets written, it shouldn't be a problem if we don't close the VersionedHDF5File (which is exactly what we were doing before). So yeah, I should maybe remove it.

versioned_hdf5/api.py
```diff
@@ -150,3 +156,25 @@ def stage_version(self, version_name: str, prev_version=None,
...
+            self._closed = True
+
+    def __repr__(self):
+        """
+        Prints friendly status information.
+
+        These messages are intended to be similar to h5py messages.
+        """
+        if self._closed:
+            return "<Closed VersionedHDF5File>"
```
scopatz, line 177: We could probably provide some of the same metadata in this string as below

melissawm: If the `VersionedHDF5File` object is already closed, it doesn't have any metada associated with it anymore (so the object file and metadata is no longer reachable). Is there anything specifically you think might be good to include in this string?

scopatz: oh ok.... nm then

scopatz:
LGTM!

PR #86, opened 2020-06-24
Adds timestamps as attributes of version groups.

### LINKED ISSUE #9: Add timestamps to the versions

melissawm:
I'm not sure the attributes is the right place to store the timestamps, let me know if you think there's a better choice.

Closes #9

### DIFF
--- a/versioned_hdf5/versions.py
+++ b/versioned_hdf5/versions.py
@@ -2,6 +2,7 @@ from uuid import uuid4
 from collections import defaultdict
 
 from h5py import Dataset, Group
+import datetime
 
 from .backend import write_dataset, write_dataset_chunks, create_virtual_dataset
 from .wrappers import InMemoryGroup, InMemoryDataset, InMemoryArrayDataset
@@ -98,6 +99,8 @@ def commit_version(version_group, datasets, *,
         create_virtual_dataset(f, version_name, name, slices, attrs=attrs,
                                fillvalue=fillvalue)
     version_group.attrs['committed'] = True
+    version_group.attrs['timestamp'] = str(datetime.datetime.now(datetime.timezone.utc))
+
 
 def delete_version(f, version_name, new_current=None):
     """
--- a/versioned_hdf5/versions.py
+++ b/versioned_hdf5/versions.py
@@ -8,6 +8,8 @@ import numpy as np
 from .backend import write_dataset, write_dataset_chunks, create_virtual_dataset
 from .wrappers import InMemoryGroup, InMemoryDataset, InMemoryArrayDataset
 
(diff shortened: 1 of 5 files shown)


asmeurer:
We should also make it so that `VersionedHDF5File.__getitem__` can lookup a version by timestamp (just searching all the versions until it finds the one it needs should be fine). We might also want to add a way to specify your own timestamp instead of using datetime.now(). At least that would make testing easier.

asmeurer:
Also storing it on an attr on the group is what I would do.

scopatz:
Also, this has conflicts now

melissawm:
I reworded the quickstart guide to include all the changes to the API. For now I'm only considering UTC in order to avoid conflicting version keys in case two versions are added in different timezones. However, since we can manually add timestamps, this can still happen. To solve this, we can either
- Implement timestamps in `create_version_group`: this has the disadvantage of assigning the timestamp before the version is actually commited.
- Implement timestamps in `commit_version` (current status): this has the disadvantage that inside `commit_version` there's no easy way to access the other previously committed timestamps to check for duplicates.

If you have any ideas on this let me know. For now there is no check for duplicate timestamps.

asmeurer:
> If you have any ideas on this let me know. For now there is no check for duplicate timestamps.

I think it's fine as it is. We can open a long term issue for this. Duplicate versions would be quite rare. The current code isn't designed to work concurrently, so it wouldn't be very likely to happen on accident. And the timestamps don't actually represent the order of the versions (that is done by the `prev_version` pointer), so the worst that will happen if there is a duplicate is that an access will get the first one it finds without warning.

melissawm:
In that case, this is ready for review.

versioned_hdf5/versions.py
```diff
@@ -98,6 +101,19 @@ def commit_version(version_group, datasets, *,
...
+            if timestamp.tzinfo != datetime.timezone.utc:
+                raise ValueError("timestamp must be in UTC")
+            version_group.attrs['timestamp'] = timestamp.strftime("%Y-%m-%d %H:%M:%S.%f%z")
+        elif isinstance(timestamp, np.datetime64):
+            version_group.attrs['timestamp'] = f"{timestamp.astype(datetime.datetime)}+0000"
+        else:
+            raise TypeError("timestamp data must be either a datetime.datetime or numpy.datetime64 object")
+    else:
+        ts = datetime.datetime.now(datetime.timezone.utc)
+        version_group.attrs['timestamp'] = ts.strftime("%Y-%m-%d %H:%M:%S.%f%z")
```
scopatz, line 115: Maybe the format string could be defined globally somewhere in its own variable?

scopatz:
@asmeurer - any comments?

PR #90, opened 2020-07-10
Support for storing binary data as void

### LINKED ISSUE #69: fillvalue and void dtypes
We try to write empty binary data to HDF5 which now fails with the changes in #60 :
```
import h5py
import numpy as np

    vf = VersionedHDF5File(f)
        sv['bar'] = [np.void(b' ')]
```
in
```
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/wrappers.py", line 338, in __init__
    self.fillvalue = fillvalue or array.dtype.type()
TypeError: void() takes exactly 1 argument (0 given)
```

ArvidJB:
This fixes #69. Storing void data is kind of weirdly supported by h5py
and we seem to use this particular corner case.

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -1064,3 +1064,18 @@ def test_multidimsional():
         version2 = file['version2']
         assert version2['test_data'][0, 1] == 2
         assert_equal(version2['test_data'][()], data2)
+
+def test_store_binary_as_void():
+    with setup() as f:
+        vf = VersionedHDF5File(f)
+        with vf.stage_version('version1') as sv:
+            sv['test_store_binary_data'] = [np.void(b'1111')]
+
+        version1 = vf['version1']
+        assert_equal(version1['test_store_binary_data'][0], np.void(b'1111'))
+
+        with vf.stage_version('version2') as sv:
+            sv['test_store_binary_data'][:] = [np.void(b'1234567890')]
+
+        version2 = vf['version2']
+        assert_equal(version2['test_store_binary_data'][0], np.void(b'1234'))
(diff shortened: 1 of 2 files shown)


asmeurer:
So this means that void datasets cannot be resized?

asmeurer:
It looks like they can and it uses nulls as the default fillvalue. 

```py
>>> f = h5py.File('test.hdf5', 'w')
>>> f.create_dataset('data', chunks=(1,), maxshape=(None,), data = [np.void(b'1111')])
<HDF5 dataset "data": shape (1,), type "|V4">
>>> f['data'].resize((2,))
>>> f['data'][:]
array([b'\x31\x31\x31\x31', b'\x00\x00\x00\x00'], dtype='|V4')
```

So maybe instead of using `dtype()` to get the default fillvalue we should use `np.zeros((1,), dtype=self.dtype)[0]`. 

```py
>>> np.zeros((1,), f['data'].dtype)[0]
void(b'\x00\x00\x00\x00')
```

asmeurer:
Or just `np.zeros((), dtype=dtype)`.

asmeurer:
I pushed that up as an alternate fix. It still passes your test.

ArvidJB:
That's probably even better, let me try your suggestion.

ArvidJB:
You probably want
```np.zeros((), array.dtype)[()]```
though, no? Otherwise the fillvalue is an array.

asmeurer:
Shape `()` arrays are usually interchangeable for scalar objects, though there can be subtle corner cases. But the scalar would match the h5py behavior, so I've modified it like you suggested.

melissawm:
Looks good, can we merge?

asmeurer:
Sorry I mean to merge this earlier.

Issue #24, opened 2020-03-27
Support writing strings in versioned file (shape 0 arrays)


ArvidJB:
In h5py you can write strings/bytes directly to a file:
```
In [1]: import h5py

In [2]: with h5py.File('foo.h5', 'w') as f:
   ...:     f['bar'] = b'baz'
   ...:

In [3]: with h5py.File('foo.h5', 'r') as f:
   ...:     print(f['bar'][()])
   ...:
b'baz'
```
Note that the "shape" is an empty tuple.
We should add support for this in versioned-hdf5.

asmeurer:
Assigning bytes strings should now work, just as with #26, since we now use np.asarray. The issue here though is that we don't support shape 0 arrays. This should be considered as part of multidimensional array support, since most places in versioned-hdf5 currently have single dimension hard-coded.

melissawm:
An update on this issue and PR #96 :

While trying to investigate the failure to create a second version as pointed out in the comments of PR #96, I found out that there is a HDF5 bug related to scalar datasets: https://github.com/h5py/h5py/issues/1409

The origin of the problem was the VirtualSource/VirtualLayout linking; since both were created with the same shape as the data, in this case a shape-0 array, the virtual dataset was not being created properly. As a workaround I increased the shapes for the VirtualSource/VirtualLayout objects, while keeping the original data with shape 0. This works if we access the data via Versioned HDF5, but not via h5py. This is in #96 

As I understand it, here's what's happening:
- Creating a shape-0 **dataset** works in h5py (see above)
- Creating a shape-0 **virtual dataset** does not work in h5py (see https://github.com/h5py/h5py/issues/1409)


Example:

```python
In [1]: import h5py; from versioned_hdf5 import VersionedHDF5File                                   

In [2]: f = h5py.File('test.h5', 'w'); file = VersionedHDF5File(f)                                  

In [3]: with file.stage_version('version1') as group: 
   ...:     group['scalar_ds'] = b'baz' 
   ...:                                                                                             

In [4]: file['version1']['scalar_ds']                                                               
Out[4]: <InMemoryArrayDataset "scalar_ds": shape (), type "|S3">

In [5]: file['version1']['scalar_ds'][()]                                                           
Out[5]: b'baz'

In [6]: f['_version_data/versions']['version1']['scalar_ds']                                        
Out[6]: <HDF5 dataset "scalar_ds": shape (1,), type "|S3">

[...]

PR #96, opened 2020-07-15
Added string datasets support (shape-0 arrays)

melissawm:
One difference between this and h5py is that 
```
f = h5py.File('file.h5', 'w')
f['bar'] = b'bar'
```
makes `f['bar']` into a `<HDF5 dataset "bar": shape (), type "|O">` (the dataset has dtype object) while 
```
vf = VersionedHDF5File(f)
with vf.stage_version('version1') as group:
    group['string_ds'] = b'baz'
```
results in a dataset with dtype S.


asmeurer:
I'm afraid this will need to be redone now that #72 is merged. Hopefully it isn't too hard. Let me know if you have any issues. 

It looks like this fails if you try to create a second version, because it expects the dataset to be a virtual dataset.

```py
>>> with file.stage_version('version1') as group:
...     group['scalar_ds'] = 0
>>> with file.stage_version('version2') as group:
...     print(group['scalar_ds'])
  File "<stdin>", line 1, in <module>
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "./versioned_hdf5/api.py", line 160, in stage_version
    prev_version=prev_version)
  File "./versioned_hdf5/versions.py", line 46, in create_version_group
    prev_group.visititems(_get)
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 565, in visititems
    return h5o.visit(self.id, proxy)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5o.pyx", line 355, in h5py.h5o.visit
  File "h5py/defs.pyx", line 1641, in h5py.defs.H5Ovisit_by_name
  File "h5py/h5o.pyx", line 302, in h5py.h5o.cb_obj_simple
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 564, in proxy
    return func(name, self[name])
  File "./versioned_hdf5/versions.py", line 40, in _get
    group[name] = item
  File "./versioned_hdf5/wrappers.py", line 91, in __setitem__
    self._data[name] = InMemoryDataset(obj.id)
  File "./versioned_hdf5/wrappers.py", line 319, in __init__
    super().__init__(InMemoryDatasetID(bind.id), **kwargs)
  File "./versioned_hdf5/wrappers.py", line 435, in __init__
    for j in range(dcpl.get_virtual_count())]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5p.pyx", line 836, in h5py.h5p.PropDCID.get_virtual_count
ValueError: Not a virtual storage layout (not a virtual storage layout)
```

Also it would be good to add tests for other types of scalars than just byte strings.

melissawm:
While trying to investigate the failure to create a second version as pointed out in the comments of PR #96, I found out that there is a HDF5 bug related to scalar datasets: https://github.com/h5py/h5py/issues/1409 

See #24  for details.  This is a partial workaround, and feedback is welcome!

asmeurer:
Is this ready to review? How does it workaround the problem?

melissawm:
I'm not sure it's ready, but I'd like some feedback. I'm not sure if it solves the problem entirely. Right now, due to the HDF5 bug, the situation is as follows:

```python
In [1]: import h5py; from versioned_hdf5 import VersionedHDF5File                                   

In [2]: f = h5py.File('test.h5', 'w'); file = VersionedHDF5File(f)                                  

In [3]: with file.stage_version('version1') as group: 
   ...:     group['scalar_ds'] = b'baz' 
   ...:                                                                                             

In [4]: file['version1']['scalar_ds']                                                               
Out[4]: <InMemoryArrayDataset "scalar_ds": shape (), type "|S3">

In [5]: file['version1']['scalar_ds'][()]                                                           
Out[5]: b'baz'
```
This works as expected; the shape of the dataset is `()`, but the byte-string has been converted to a NumPy 'S' dtype.

However:
```python
In [6]: f['_version_data/versions']['version1']['scalar_ds']                                        
Out[6]: <HDF5 dataset "scalar_ds": shape (1,), type "|S3">
```
The virtual dataset has shape `(1,)`, as otherwise we couldn't link the VirtualLayout to the VirtualSource. Also, if we try to access the 0-shaped array directly using the HDF5 API, we get 
```python
In [7]: f['_version_data/versions']['version1']['scalar_ds'][()]                                    
Floating point exception (core dumped)
```
This is the bug I mentioned above.

asmeurer:
I think it will be more robust to just copy scalar datasets, meaning don't make them virtual at all. There are a few places in the code that assume a dataset is virtual that would need to be updated.  Hopefully not too many. If the bug in the upstream is ever fixed, we can modify it to use virtual datasets again. 

The only concern is if this would lead to too large of a memory footprint, since every version would have every scalar dataset copied in the file. It sounds like DEShaw is using large byte strings, so this could be an issue. If so, then I think what we have here is fine. We already specify that you should only use the versioned_hdf5 API to access versioned files. We just need to be careful whenever we read through virtual datasets in library code to check if they are scalar.

asmeurer:
As discussed in the meeting today, this should be good enough for now.

PR #107, opened 2020-07-27
Fix chunks check

### LINKED ISSUE #103: Bug in `set_chunks` which leads to the wrong chunks being used:
This still seems to be an issue after merging #72:
```
import h5py
import numpy as np

    vf = VersionedHDF5File(f)
        data_group = sv.create_group('data')
        data_group.create_dataset('bar', data=np.arange(4))

    vf = VersionedHDF5File(f)
        data_group = sv['data']
        data_group.create_dataset('props/1/bar', data=np.arange(0, 4, 2))
```
which complains about
```
  File "/u/bessen/opensource/versioned-hdf5/set_chunks_dict_bug.py", line 15, in <module>
    data_group.create_dataset('props/1/bar', data=np.arange(0, 4, 2))
    next(self.gen)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/api.py", line 170, in stage_version
    timestamp=timestamp)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/versions.py", line 96, in commit_version
    raise NotImplementedError("Specifying chunk size with dict data")
NotImplementedError: Specifying chunk size with dict data
```

_Originally posted by @ArvidJB in https://github.com/deshaw/versioned-hdf5/pull/72#issuecomment-659119626_

asmeurer:
Fixes #103.

### DIFF
--- a/versioned_hdf5/tests/helpers.py
+++ b/versioned_hdf5/tests/helpers.py
@@ -3,10 +3,12 @@ import h5py
 from ..backend import initialize
 
 # TODO: Use a fixture for this
-def setup(file_name='test.hdf5', name=None, version_name=None):
+def setup(file_name='test.hdf5', name=None, version_name=None, init=True):
     # TODO: Use a temporary directory
-    f = h5py.File(file_name, 'w')
-    initialize(f)
+    mode = 'w' if init else 'r+'
+    f = h5py.File(file_name, mode)
+    if init:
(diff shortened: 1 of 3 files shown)


asmeurer:
Actually it looks like there was a bug here, and https://github.com/deshaw/versioned-hdf5/pull/104 is the right fix.

PR #108, opened 2020-07-27
Add a size attribute to InMemoryArrayDataset

### LINKED ISSUE #105: Add "size" property to InMemoryArrayDataset
This code fails
```
    vf = VersionedHDF5File(f)
        x = sv.create_dataset('bar', data=np.arange(4))
        x.size
```
with
```
AttributeError: 'InMemoryArrayDataset' object has no attribute 'size'
```
We should add a "size" property like
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html

asmeurer:
Fixes #105.


scopatz:
Thanks!

Issue #111, opened 2020-07-28
Variable length string support


ArvidJB:
We use the pre-h5py-2.9 method to store variable length strings, http://docs.h5py.org/en/2.8.0/strings.html :
```
In [21]: with h5py.File('foo.h5', 'w') as f:
    ...:     dt = h5py.special_dtype(vlen=bytes)
    ...:     dset = f.create_dataset("name", (10,), dtype=dt, data=np.full(10, 'hello world', dtype='O'))
    ...:
```
but this does not work with versioned_hdf5:
```
In [23]: with h5py.File('foo.h5', 'w') as f:
    ...:     vf = VersionedHDF5File(f)
    ...:     with vf.stage_version('0') as sv:
    ...:         dt = h5py.special_dtype(vlen=bytes)
    ...:         dset = sv.create_dataset("name", shape=(10,), dtype=dt, data=np.full(10, 'hello world', d
    ...: type='O'))
    ...:
```
which fails with
```
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
```

asmeurer:
We aren't passing the detype around properly. But an issue here is how to handle the fillvalue. The current logic to generate a default fillvalue fails for `dt`.

asmeurer:
Partial fix at https://github.com/deshaw/versioned-hdf5/pull/113

asmeurer:
Do we need to support h5py 2.8 as well? My understanding was that 2.9 was OK. #99

ArvidJB:
We are using 2.9. But our code was written pre 2.9 so it uses the old h5py way to support variable length strings.

asmeurer:
I guess the fillvalue issue is an issue with h5py https://github.com/h5py/h5py/issues/941. h5py sets the fillvalue to b'' automatically if none is passed and that seems to work, so we can work around this. Setting a fillvalue of something other than '' won't work, but I don't think that presently works in base h5py either.

asmeurer:
Fix at https://github.com/deshaw/versioned-hdf5/pull/115. There will be issues if there are other object dtypes supported by h5py, as we need to support each type manually.

Issue #25, opened 2020-03-27
Support attrs on versioned Datasets


ArvidJB:
It looks like `attrs` is not supported on versioned Datasets?
```
In [10]: with h5py.File('foo.h5', 'w') as f:
    ...:     vf = VersionedHDF5File(f)
    ...:     with vf.stage_version('0') as sv:
    ...:         sv.create_dataset('bar', data=np.array([1, 2, 3]))
    ...:         sv['bar'].attrs['bla'] = 123
    ...:

In [11]: with h5py.File('foo.h5', 'r') as f:
    ...:     vf = VersionedHDF5File(f)
    ...:     v0 = vf['0']
    ...:     print(v0['bar'][:])
    ...:     print(v0['bar'].attrs['bla'])
    ...:
[1 2 3]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-11-51e42db215d4> in <module>
      3     v0 = vf['0']
      4     print(v0['bar'][:])
----> 5     print(v0['bar'].attrs['bla'])
      6

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

/usr/local/python/python-3.7/std/lib64/python3.7/site-packages/h5py/_hl/attrs.py in __getitem__(self, name)
     58         """ Read the value of an attribute.
     59         """
---> 60         attr = h5a.open(self._id, self._e(name))
     61
     62         if is_empty_dataspace(attr):

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/h5a.pyx in h5py.h5a.open()

KeyError: "Can't open attribute (can't locate attribute: 'bla')"
```

scopatz:
@asmeurer - I think this just needs to be passed through the API

asmeurer:
Yes. I think we should just duplicate all attrs on all the versions of the dataset, i.e., not try to do de-duplication, as I think that would be a complicated optimization and attributes that are large cause other issues in hdf5 anyway.

ArvidJB:
I think it's fair to assume that attrs are always going to be "small" (at least for us) and I agree with @asmeurer: we can just store another copy of them with each version even if they did not change.

ArvidJB:
Note that with the latest code this fails even earlier because the "dataset" that gets created is now just a numpy array:
```
[...]

PR #113, opened 2020-07-28
Keep track of the dtype when creating a dataset

asmeurer:
h5py has its own custom object dtype objects that are used to support strings.
See issue #111.

Note that this does not yet fix that issue, because there is a problem trying
to create a fillvalue for an object dtype. I'm not sure how to handle that.


melissawm:
This is related to #25 and #96.

PR #115, opened 2020-07-30
Add support for variable length string dtypes

### LINKED ISSUE #111: Variable length string support

asmeurer:
We need to manually handle the fillvalue, as there are some bugs in h5py that
prevent it from working. Also, the default fillvalue can show up as None if we
aren't careful, since it can look like an object dtype.

Fixes #111.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -46,6 +46,14 @@ def create_base_dataset(f, name, *, shape=None, data=None, dtype=None,
         else:
             raise NotImplementedError("chunks must be specified for multi-dimensional datasets")
     group = f['_version_data'].create_group(name)
+    if dtype.metadata and ('vlen' in dtype.metadata or 'h5py_encoding' in dtype.metadata):
+        # h5py string dtype
+        # (https://h5py.readthedocs.io/en/2.10.0/strings.html). Setting the
+        # fillvalue in this case doesn't work
+        # (https://github.com/h5py/h5py/issues/941).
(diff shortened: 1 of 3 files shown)


melissawm:
@asmeurer I think something went wrong with the CI, should we retrigger?

scopatz:
@asmeurer @melissawm - It seems like this has stalled out

scopatz:
The CI that is

asmeurer:
In the future, you can just push the branch up to the repo, as I have done, or else push up a dummy commit. Sometimes Travis doesn't trigger for whatever reason. Sometimes it's because the .travis.yml file is broken, but in this case, there were no changes to that file.

PR #116, opened 2020-08-03
Fixes docs build.

melissawm:


asmeurer:
We should include the changelog in the docs.

PR #124, opened 2020-08-12
Added fixtures and temporary dirs for tests.

### LINKED ISSUE #88: Make the tests use a temporary directory and a fixture

melissawm:
Including some configs to make fixtures more reusable.

Fixes #88

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -508,6 +508,20 @@ def test_resize(vfile):
     assert_equal(group['data'][:], 1.0)
 
 
+def test_resize_unaligned(vfile):
+    ds_name = 'test_resize_unaligned'
+    with vfile.stage_version('0') as group:
+        group.create_dataset(ds_name, data=np.arange(1000))
+
+    for i in range(1, 10):
+        with vfile.stage_version(str(i)) as group:
+            l = len(group[ds_name])
+            assert_equal(group[ds_name][:], np.arange(i * 1000))
+            group[ds_name].resize((l + 1000,))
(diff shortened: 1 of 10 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -296,6 +296,7 @@ def _make_new_dset(shape=None, dtype=None, data=None, chunks=None,
                   track_order=None, dcpl=None):
     """ Return a new low-level dataset identifier """
 
+    print(f"shape = {shape}")
```
asmeurer, line 299: I think this shouldn't be here.

versioned_hdf5/tests/test_api.py
```diff
@@ -10,1037 +10,925 @@
...
-        group = file['resize_after_create_smaller']
         assert group['data'].shape == (DEFAULT_CHUNK_SIZE - 4,)
         assert_equal(group['data'][:], 1.0)
 
-        file.close()
+    group = vfile['resize_after_create_smaller']
+    assert group['data'].shape == (DEFAULT_CHUNK_SIZE - 4,)
+    assert_equal(group['data'][:], 1.0)
 
-def test_resize_unaligned():
```
asmeurer, line 564: This test was deleted.

asmeurer:
Do we still need the `setup()` helper function? I see it's still used in some places.

melissawm:
> Do we still need the `setup()` helper function? I see it's still used in some places.

The difficulty is that some tests re-run the setup routine several times, and this can't be done with a fixture, which is run once before the test. I couldn't figure a way to remove the setup function from those.

scopatz:
@melissawm - is this ready for review?

PR #130, opened 2020-09-09
Fix support for empty datasets

### LINKED ISSUE #98: Writing empty dataset fails
```
import h5py
import numpy as np

    vf = VersionedHDF5File(f)
        sv['bar'] = np.arange(0)
```
fails with
```
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/backend.py", line 152, in create_virtual_dataset
    shape = (chunk_size*(len(slices) - 1) + slices[-1].stop - slices[-1].start,) + chunks[1:]
IndexError: list index out of range
```
in master and with
```
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/backend.py", line 155, in create_virtual_dataset
    shape = tuple([max(c.args[i].stop for c in slices) for i in range(len(chunks))])
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/backend.py", line 155, in <listcomp>
    shape = tuple([max(c.args[i].stop for c in slices) for i in range(len(chunks))])
ValueError: max() arg is an empty sequence
```
in asmeurer/as_subindex.

asmeurer:
We need to make sure we have at least one dummy entry in the data_dict so we
can infer the shape.

There is a bug with getting the virtual dataset sources for empty dataset in
h5py, so we have to work around it.

Fixes #98.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -124,7 +124,7 @@ def write_dataset_chunks(f, name, data_dict):
     chunk_size = chunks[0]
 
     shape = tuple(max(c.args[i].stop for c in data_dict) for i in
-    range(len(chunks)))
+                  range(len(chunks)))
     all_chunks = list(split_chunks(shape, chunks))
     for c in all_chunks:
         if c not in data_dict:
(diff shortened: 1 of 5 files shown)


scopatz:
CI is failing for some reason

asmeurer:
Yeah, I didn't check the full test suite, and there are some failures. I think they will be easy to fix.

scopatz:
Thanks!

PR #135, opened 2020-09-11
Allow deleting groups between versions

### LINKED ISSUE #131: Deleting datasets
We need to support deleting datasets/groups and recreating them in later versions.

asmeurer:
Deleting datasets was already supported, but not tested.

This works as long as the dimensionality or dtype does not change between
versions. Changing the dtype would be possible, but would require some work.
Changing the dimensionality would be more difficult.

Fixes #131.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -82,7 +82,13 @@ def write_dataset(f, name, data, chunks=None, compression=None,
     if compression or compression_opts:
         raise ValueError("Compression options can only be specified for the first version of a dataset")
     if fillvalue is not None and fillvalue != ds.fillvalue:
-        raise ValueError(f"fillvalues do not match ({fillvalue} != {ds.fillvalue})")
+        dtype = ds.dtype
+        if dtype.metadata and ('vlen' in dtype.metadata or 'h5py_encoding' in dtype.metadata):
+            # Variable length string dtype. The ds.fillvalue will be None in
+            # this case (see create_virtual_dataset() below)
+            pass
+        else:
+            raise ValueError(f"fillvalues do not match ({fillvalue} != {ds.fillvalue})")
     if data.dtype != ds.dtype:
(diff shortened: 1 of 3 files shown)


scopatz:
Seems as though there is a conflict here now

ArvidJB:
Thanks, that was really quick. It mostly works except for strings:
```
    vf = VersionedHDF5File(f)
        dtype = h5py.special_dtype(vlen=bytes)
        sv.create_dataset('bar/baz', data=['aaa'], dtype=dtype)

    vf = VersionedHDF5File(f)
        del sv['bar']
        sv.create_dataset('bar/baz', data=['bbb'], dtype=dtype)
```
gives me the following error
```
  File "/u/bessen/opensource/versioned-hdf5/bytes_array_fillvalue.py", line 15, in <module>
    sv.create_dataset('bar/baz', data=['bbb'], dtype=dtype)
    next(self.gen)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/api.py", line 171, in stage_version
    timestamp=timestamp)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/versions.py", line 102, in commit_version
    fillvalue=fillvalue)
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/backend.py", line 85, in write_dataset
    raise ValueError(f"fillvalues do not match ({fillvalue} != {ds.fillvalue})")
ValueError: fillvalues do not match (b'' != None)
```

asmeurer:
It looks like this bug is unrelated to deleting (it happens on master too if you remove the `del` line and use `sv['bar/baz'] = ['bbb']`). Actually, deleting a group or dataset and recreating it in the same `stage_version` block should have no effect and ought to work even in master.

asmeurer:
I guess in that case, you don't actually specify the fillvalue, which confuses it. You can use `sv['bar/baz'][()] = ['bbb']`, which does work.

asmeurer:
I pushed a fix for it.

scopatz:
Thanks

PR #136, opened 2020-09-11
Fix automatic creation of groups on setitem

asmeurer:
This matches the h5py behavior.


scopatz:
Seems as though there is a conflict

scopatz:
Thanks for this @asmeurer. Can we disable the inline code-cov, it is pretty annoying.

scopatz:
And also seems wrong.

asmeurer:
Oof, when did it start doing that?

scopatz:
I don't know, this is the first time I have seen it.

PR #139, opened 2020-09-14
Make sure scalar indices give scalar objects

### LINKED ISSUE #137: Return element instead of shape () array
`h5py` seems to have some special case for "single_element" `__getitem__` calls which `versioned_hdf5` currently does not have.
This is some code which currently fails with
```TypeError: len() of unsized object```
but which would work in `h5py`:
```
    vf = VersionedHDF5File(f)
        dtype = h5py.special_dtype(vlen=bytes)
        sv.create_dataset('bar', data=np.array(['aaa'], dtype='O'), dtype=dtype)

    vf = VersionedHDF5File(f)
    print(len(vf['0']['bar'][0]))
```

asmeurer:
Fixes #137.


scopatz:
Thanks!

PR #144, opened 2020-10-21
Continuing work on sparse datasets

asmeurer:
#63 was merged prematurely, so this continues that work.

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -150,6 +150,8 @@ class VersionedHDF5File:
         `prev_version` for any future `stage_version` call.
 
         """
+        if self._closed:
+            raise ValueError("File is closed")
         old_current = self.current_version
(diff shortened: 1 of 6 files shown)


asmeurer:
I have it so that sparse data works, but only if you close and reopen the file before writing any non-fillvalue data. However, full support should not be much more work. I also haven't added any tests yet.

asmeurer:
This is ready for testing and review. I think it might be a good idea to look into making the necessary improvements in ndindex to improve the performance here (currently any index on a dataset naively iterates over all chunks to see which ones contain the index). The performance is serviceable for the dataset shapes/chunk sizes described at https://github.com/deshaw/versioned-hdf5/pull/63#issuecomment-709315447, but I think it could still be noticable.

scopatz:
CI seems to be failing

scopatz:
I think this looks reasonable.

PR #147, opened 2020-11-10
Store the chunks on an attr of the virtual dataset

### LINKED ISSUE #145: No chunks on virtual dataset?
To work around the missing Python 2 support (see #10) we decided to replicate the reading logic (for the current version only) in Python 2. This works nicely, except that the virtual dataset has no `chunks` property?
```
    vf = VersionedHDF5File(f)
    with vf.stage_version('v1') as sv:
        sv.create_dataset('bar', data=np.arange(5), chunks=(10,))

    vf = VersionedHDF5File(f)
    cv = vf[vf.current_version]
    bar = cv['bar']
    assert np.all(bar[:] == np.arange(5))
    assert bar.chunks == (10,)

    versions = f['_version_data/versions']
    bar = versions[versions.attrs['current_version']]['bar']
    assert np.all(bar[:] == np.arange(5))
    assert bar.chunks == (10,)
```
Is that something that's easy to add? If not, I can add another workaround for that.

asmeurer:
Virtual datasets don't have the chunks property, so this is the only way to
access it when reading it from raw HDF5.

Fixes #145.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -212,4 +212,5 @@ def create_virtual_dataset(f, version_name, name, shape, slices, attrs=None, fil
         for k, v in attrs.items():
             virtual_data.attrs[k] = v
     virtual_data.attrs['raw_data'] = raw_data.name
+    virtual_data.attrs['chunks'] = raw_data.chunks
     return virtual_data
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM!

PR #148, opened 2020-11-10
Fixing things for h5py 3

asmeurer:
h5py 3.10 is out now, and there are several test failures. I've also added a CI build for Python 3.7 + h5py 2.10, which as I understand it is the configuration currently used internally at DEShaw (correct me if this is wrong), and I added CI tests for Python 3.9.

PR #150, opened 2020-11-16
Pin h5py to version 2 for now

asmeurer:
Version 3 support is being worked on in #148, but will take some time to get
working, and we want to do a release before then.


scopatz:
LGTM

Issue #146, opened 2020-11-10
Version replaying


asmeurer:
We need a way to "replay" versions, i.e., to reconstruct the raw data for a dataset and rebuild the data in each version. This will enable several features

- Changing certain metadata about a dataset, which currently cannot be changed across versions, including `ndim`, `chunks`, and `dtype` https://github.com/deshaw/versioned-hdf5/issues/110
- Support for deleting a version. Currently we can delete the version group, but it would not remove the data for that version from the raw data. https://github.com/deshaw/versioned-hdf5/issues/143
- It opens the possibility of doing some tricks like inserting rows without actually rewriting all the data, by leaving some data in the raw data empty https://github.com/deshaw/versioned-hdf5/issues/47. The issue with this is that you need a way to garbage collect the empty data after a while, or else it becomes more inefficient than just doing it the naive way.

asmeurer:
For changing metadata, there are two points. 

1. If we want to only change the metadata for new versions, but keep old versions as they are, I think this issue isn't what we need. Rather, we need some way to point the new dataset to a different raw dataset. Currently the raw dataset is referenced by name, but I think this wouldn't be hard to change, as we already are storing the path to the raw dataset in the metadata ever since #144 

2. If we want to change it for all old versions, this is what is needed. This is a potentially destructive action, so it may be worth having some checks that the new metadata isn't lossy against the old without some sort of force flag (e.g., the new `ndim` should be larger, the new dtype should cast from the old losslessly, etc.).

ArvidJB:
Sounds good. Let's make this as simple as possible. It would be great if we could have some API which takes a callback and then calls it for each version to write a new HDF5 file:
* Input would be the `InMemoryGroup` for that version (including attrs)
* Output would be `None` (if we want to skip this version) or a new `InMemoryGroup` with added/removed/modified datasets/attrs

Do you think this would work?

asmeurer:
I hadn't considered doing everything in a new file. That would take care of the atomicity concerns, though it also means we would need to copy everything.

asmeurer:
Version replaying is now implemented in versioned_hdf5/replay.py, and the modify_metadata and delete_version functions in particular. The insertion idea has not yet been implemented, but we can use #47 to track that, so I am closing this issue.

PR #152, opened 2020-11-19
Version replaying

asmeurer:
This is the work for #146 (still in progress).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,7 +12,7 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
 
 # -- Project information -----------------------------------------------------
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -68,12 +68,13 @@ def create_base_dataset(f, name, *, shape=None, data=None, dtype=None,
     dataset.attrs['chunks'] = chunks
     return write_dataset(f, name, data, chunks=chunks)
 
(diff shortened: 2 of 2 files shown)


asmeurer:
I'm using the callback idea from https://github.com/deshaw/versioned-hdf5/issues/146#issuecomment-724994295. It works at the dataset level, rather than the group level, as this simplifies the case where you want to just change a single dataset. Certain callback patterns will be common enough that we can add them as helpers. 

For example, to change the dtype of a dataset, you can do something like

```py
from versioned_hdf5.backend import tmp_group, recreate_dataset
tmp = tmp_group(f)

tmpgroup = InMemoryGroup(tmp.create_group('tmp_group').id)

def callback(dataset, version):
    _, name = pp.split(dataset.name) # This will probably change with issue #154
    return dataset.as_dtype(name, np.int64, tmpgroup)

recreate_dataset(f, dataset_name, tmp, callback=callback)
```

(note that `as_dtype` is a new method on versioned_hdf5 dataset objects that returns a new dataset with a new dtype). The callback can also return None to mean the dataset should be deleted for the given version.

What remains to be implemented is

- Logic to automatically create the temporary group as above, and more importantly, to merge the recreated data back into the main _version_data group. 
- A function to run recreate_dataset on every dataset in every version (so e.g., you can provide a callback that causes a version to be deleted).

asmeurer:
This is ready for initial review. I've added the top-level functions `delete_version` and `modify_metadata`. I'm still unsure about the API, so feel free to suggest something better. There are also some useful internal helper functions, notably `recreate_dataset`, which allow doing arbitrary replay actions with a callback.

asmeurer:
Just published version 1.3 with these changes.

PR #153, opened 2020-11-23
Fix creating a completely empty sparse dataset

asmeurer:


docs/conf.py
```diff
@@ -12,7 +12,7 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
```
scopatz, line 15: Wouldn't `os.path.join` be better as  `os.path.dirname`?

asmeurer: I don't really follow this.

PR #172, opened 2021-04-08
Add a link to the benchmarks suite in the docs

asmeurer:


asmeurer:
Ah, bad syntax. I don't know why we're still using RST for this project.

PR #176, opened 2021-04-15
Remove codecov

asmeurer:
They recently had a security incident.
https://about.codecov.io/security-update/. We haven't actually been making use
of codecov, so it's best to just remove it. We can always get coverage output
by manually running the tests with --cov.

The COVERAGE build on Travis was separate from the DOCS build with the secure
upload token, so I don't think the codecov security incident could have leaked
anything important.


asmeurer:
I don't seem to have the permissions necessary to remove codecov here https://github.com/deshaw/versioned-hdf5/settings/installations

asmeurer:
Codecov was disabled in the deshaw organization.

PR #180, opened 2021-04-19
Store the chunks directly on InMemoryArrayDataset and make sure it is maintained in DatasetWrapper

### LINKED ISSUE #179: chunks is lost when entire Dataset is overwritten
If the slice we are writing to is the entire dataset then the `chunks` attribute is lost:
```
    vf = VersionedHDF5File(f)
        bar = sv.create_group('bar')
        bar.create_dataset('baz', data=[1, 2, 3], chunks=(2,))

    vf = VersionedHDF5File(f)
        bar = sv['bar']
        assert bar['baz'].chunks == (2,)
        bar['baz'][0:3] = [4, 5, 6]
        assert bar['baz'].chunks == (2,)
```
This seems to have been introduced in 1.2.4, I think?
Can you take a look at all the attributes after `DatasetWrapper.__setitem__` is called? This seems to be a very similar issue to #174 174.

asmeurer:
Fixes #179.

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -1686,10 +1686,11 @@ def test_empty_dataset_str_dtype(vfile):
 
 def test_datasetwrapper(vfile):
     with vfile.stage_version('r0') as sv:
-        sv.create_dataset('bar', data=[1, 2, 3])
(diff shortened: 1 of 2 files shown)


ArvidJB:
Thanks, can we do another release with this fix?

PR #195, opened 2021-06-18
Unpin h5py

takluyver:
It looks like PR #148 fixed the incompatibilities and resumed running the CI with h5py 3.x, so hopefully the requirement can be relaxed now.


asmeurer:
Oh I forgot that we had a pin in setup.py. Yes, that should have been removed back when #148 was merged. We do still need to support 2.10 in this repo, although work is being done to upgrade to h5py 3.

Issue #201, opened 2021-07-31
Selecting with bool mask broken?


ArvidJB:
This seems to no longer work as of the latest release:
```
from __future__ import absolute_import, division

from   versioned_hdf5           import VersionedHDF5File

import h5py
import numpy as np

with h5py.File('foo.h5', 'w') as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version('r0') as sv:
        sv.create_dataset('bar', data=[1, 2, 3], chunks=(2,))

with h5py.File('foo.h5', 'r+') as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version('r1') as sv:
        mask = np.array([True, True, False], dtype='bool')
        b = sv['bar'][mask]
```
gives the following error
```
Traceback (most recent call last):
  File "/u/bessen/opensource/versioned-hdf5/bool_mask.py", line 17, in <module>
    b = sv['bar'][mask]
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/wrappers.py", line 1107, in __getitem__
    return self.dataset.__getitem__(index)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/u/bessen/opensource/versioned-hdf5/versioned_hdf5/wrappers.py", line 644, in __getitem__
    return super().__getitem__(args)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/usr/local/python/python-3.7/std/lib64/python3.7/site-packages/h5py/_hl/dataset.py", line 573, in __getitem__
    self.id.read(mspace, fspace, arr, mtype, dxpl=self._dxpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 182, in h5py.h5d.DatasetID.read
  File "h5py/_proxy.pyx", line 130, in h5py._proxy.dset_rw
  File "h5py/_proxy.pyx", line 84, in h5py._proxy.H5PY_H5Dread
OSError: Can't read data (point selections not currently supported)
```
[...]

PR #203, opened 2021-08-02
1.3.3 release

asmeurer:
@ArvidJB I've added documentation on how to make a release. You may want to look it over and see if it looks OK.


asmeurer:
I updated the masking fix as per the discussion at #201. 1.3.3 has been released. The release docs are now live at https://deshaw.github.io/versioned-hdf5/releasing.html.

PR #206, opened 2021-09-23
Make the hashtable dataset exactly as big as it needs to be

### LINKED ISSUE #205: Empty file taking up lots of disk space
A versioned .h5 file which contains no data still takes up a lot of disk space. To reproduce this you can create a file like this:
```
In [1]: import h5py

In [2]: from versioned_hdf5 import VersionedHDF5File

In [3]: with h5py.File('data.h5', 'w') as f:
   ...:         sv['foo'] = np.array([], dtype='int')
   ...:
```
which takes about 200KB of disk space:
```
~> ls -l data.h5
-rw-rw-r-- 1 bessen bessen 210328 Sep 10 09:54 data.h5
```
It seems this is due to the `hash_table`, every other dataset does not actually end up taking any space:
```
~> h5ls -v -r data.h5
...
/_version_data/foo/hash_table Dataset {4096/Inf}
    Attribute: largest_index scalar
        Type:      native long
        Data:  0
    Location:  1:10552
    Links:     1
    Chunks:    {4096} 196608 bytes
    Storage:   196608 logical bytes, 196608 allocated bytes, 100.00% utilization
    Type:      struct {
                   "hash"             +0    [32] native unsigned char
                   "shape"            +32   [2] native long
               } 48 bytes
...
```

Is there a way to make the `hash_table` take less space?

asmeurer:
Previously it was always filled with 0s to be default_chunk_size aligned, but
this is no longer necessary now that we write to it once at the end. The
chunk_size resizing is still used for the in-memory NumPy array.

This makes the hashtables much smaller on disk for datasets without many
chunks.

This keeps the largest_index attribute for backwards compatibility with hash
tables that were created before this change (which will automatically be
resized down the next time they are written to). It also fixes the
documentation of largest_index.

Fixes #205.

### DIFF
--- a/versioned_hdf5/hashtable.py
+++ b/versioned_hdf5/hashtable.py
@@ -22,8 +22,8 @@ class Hashtable(MutableMapping):
     where setdefault will insert the hash into the table if it
     doesn't exist, and return the existing entry otherwise.
 
-    hashtable.largest_index is the largest index in the array that has slices
-    mapped to it.
+    hashtable.largest_index is the next index in the array that slices
+    should be mapped to.
 
     Note that for performance reasons, the hashtable does not write to the
     dataset until you call write() or it exit as a context manager.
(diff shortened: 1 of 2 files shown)


ArvidJB:
Looks good to me. What are the file sizes now for the example code in #205 ?

asmeurer:
Before data.h5 was 205K, with this PR it is 11K.

PR #209, opened 2021-09-30
Fix a bug with the hashtable introduced by #206

asmeurer:


asmeurer:
I've made 1.3.5 with this.

PR #214, opened 2021-12-16
Better delete_version algorithm

### LINKED ISSUE #213: Faster deleting of versions
We do occasionally clean up older versions. To do that we can use the `replay` APIs. However, this is surprisingly slow for the simple task it needs to do, probably because it reuses the generic `recreate_dataset` logic which allows arbitrary modifications of the dataset.

The example below takes about 3 minutes seconds to keep every 10th version of a dataset with 1000 versions. I think this can be sped up:
1. we can scan over all versions to figure out which raw chunks needs to be kept
2. we can then create a map from "old" raw chunk index to "new" raw chunk index
3. we then recreate the virtual datasets for the versions we want to keep by replacing the "old" raw chunk index with the "new" raw chunk index.

```
import datetime
import logging
import os
import shutil
import tempfile

import h5py
import numpy

from versioned_hdf5.replay import recreate_dataset, tmp_group, swap


def delete_versions(f, names, versions_to_delete):
    """
    Modified replay.delete_version to delete multiple versions.
    """
    if isinstance(f, VersionedHDF5File):
        f = f.f

    def callback(dataset, version_name):
        if version_name in versions_to_delete:
            return
        return dataset

    newf = tmp_group(f)

    for name in names:
        recreate_dataset(f, name, newf, callback=callback)

    swap(f, newf)

    for version in versions_to_delete:
        del f['_version_data/versions'][version]

    del newf[newf.name]


if __name__ == '__main__':
    filename = '/proj/scratch-nb-bessen/delete_versions_slow.h5'

    logging.getLogger().setLevel(logging.INFO)

    # Uncomment when running for the first time!

    # logging.info('Creating datasets...')
    # with h5py.File(filename, 'w') as f:
    #     vf = VersionedHDF5File(f)
    #     with vf.stage_version('init') as sv:
    #         sv.create_dataset('values', shape=(0, 0), dtype='float', fillvalue=numpy.nan,
    #                           chunks=(22, 100), maxshape=(None, None), compression='lzf')
    #
    # # generate some test data with around 1000 versions
    # v = 1
    # with h5py.File(filename, 'r+') as f:
    #     vf = VersionedHDF5File(f)
    #     for d in range(22):
    #         with vf.stage_version(str(v)) as sv:
    #             values_ds = sv['values']
    #             values_ds.resize((values_ds.shape[0] + 1, values_ds.shape[1] + 5000))
    #             values_ds[-1, -5000] = numpy.random.rand()
    #             v += 1
    #         for c in range(50):
    #             with vf.stage_version(str(v)) as sv:
    #                 values_ds = sv['values']
    #                 idxs = numpy.random.choice(values_ds.shape[1], 50, replace=False)
    #                 values_ds[-1, idxs] = numpy.random.rand(50)
    #                 v += 1

    tmp_name = tempfile.mktemp('.h5')
    shutil.copy2(filename, tmp_name)
    try:
        logging.info('%s: Keeping every 10th version...', datetime.datetime.now())

        # want to keep only every 10th version
        versions_to_delete = []
        with h5py.File(tmp_name, 'r') as f:
            vf = VersionedHDF5File(f)
                if i % 10 != 0:

        with h5py.File(tmp_name, 'r+') as f:
            delete_versions(f, ['values'], versions_to_delete)

        with h5py.File(tmp_name, 'r') as f:
            vf = VersionedHDF5File(f)
            logging.info('%s: Versions left: %s', datetime.datetime.now(), vf._versions.keys())
    finally:
        os.remove(tmp_name)
```

asmeurer:
This replaces delete_version with the idea described in #213. The example there now takes 45 seconds, and I think I know how to shave off an additional 15 seconds.

Fortunately, the poor performance I mentioned this morning was due to a stupid mistake on my part. I was accidentally repeating the algorithm once per version instead of just once total.

This PR also renames `delete_version()` to `delete_versions()` (but keeps the old API for backwards compatibility). 

This still needs more testing and a little bit of code cleanup. 

Fixes #213.


asmeurer:
There also appear to be some legitimate test failures here.

ArvidJB:
I just tried this, but after running `delete_versions` I get the error
```
KeyError: "Can't open attribute (can't locate attribute: 'raw_data')"
```
when I try to read the modified file. Here's what I tried
```
        vf = VersionedHDF5File(f)
            sv.create_dataset('bar', data=np.arange(10))

    for i in range(1, 11):
            vf = VersionedHDF5File(f)
            with vf.stage_version('r{}'.format(i)) as sv:
                sv['bar'] = np.arange(10 + i)

        delete_versions(f, ['r2', 'r4', 'r6'])

        vf = VersionedHDF5File(f)
        print(vf['r0']['bar'][:])
```

asmeurer:
Turns out I had a bunch of changes that I forgot to push. This should work now.

asmeurer:
I've added a benchmark for deleting and uploaded the results here https://deshaw.github.io/versioned-hdf5/benchmarks/index.html#delete_versions.TimeDeleting.time_delete?branch=new-delete&p-param1=10&p-param1=30&p-param1=50. There is unfortunately some meaningless variance in some of the runs, but the main takeaway from the graph is the order of magnitude speedup from master, then the further speedup from https://github.com/deshaw/versioned-hdf5/commit/6ebf58d6.

PR #221, opened 2022-03-29
1.3.8

asmeurer:


asmeurer:
1.3.8 is released with the fix from #218. Let me know if that doesn't fully address the issue.

PR #229, opened 2022-06-14
Set current_version in delete_versions

### LINKED ISSUE #227: Deleting last version corrupts file
When we use `delete_versions` to clean up a file and we delete the last version we then get an error the next time we read it:
```
        vf = VersionedHDF5File(f)
            sv.create_dataset('bar', data=np.arange(10))

    for i in range(1, 11):
            vf = VersionedHDF5File(f)
            with vf.stage_version('r{}'.format(i)) as sv:
                sv['bar'] = np.arange(10 + i)

        # delete versions, including last version!
        delete_versions(f, ['r2', 'r4', 'r6', 'r8', 'r10'])

        vf = VersionedHDF5File(f)
        cv = vf[vf.current_version]
        print(cv['bar'][:])
```
this raises
```
    cv = vf[vf.current_version]
  File "/usr/local/python/python-3.9/std/lib64/python3.9/site-packages/versioned_hdf5/api.py", line 130, in __getitem__
    return self.get_version_by_name(item)
  File "/usr/local/python/python-3.9/std/lib64/python3.9/site-packages/versioned_hdf5/api.py", line 101, in get_version_by_name
    raise KeyError(f"Version {version!r} not found")
KeyError: "Version 'r10' not found"
```

At the end of `delete_versions` we need to update `f['versions'].attrs['current_version']` to the last version we keep in the file.

asmeurer:
Fixes #227.

### DIFF
--- a/versioned_hdf5/replay.py
+++ b/versioned_hdf5/replay.py
@@ -319,9 +319,16 @@ def delete_versions(f, versions_to_delete):
 
     versions = version_data['versions']
 
+    if '__first_version__' in versions_to_delete:
+        raise ValueError("Cannot delete first version")
+
     for version in versions_to_delete:
         if version not in versions:
             raise ValueError(f"Version {version!r} does not exist")
+
+    current_version = versions.attrs['current_version']
+    while current_version in versions_to_delete:
+        current_version = versions[current_version].attrs['prev_version']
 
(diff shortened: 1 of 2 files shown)


asmeurer:
I'm not able to reproduce the test failure here.

PR #241, opened 2023-01-21
Fix the prev_versions in delete_versions

### LINKED ISSUE #240: delete_versions does not update "prev_version" properly, PyInf#9334
When we use `delete_versions` the "prev_version" attribute is not updated accordingly. Here's an example where we delete versions 'r1', 'r5', 'r8' and would therefore expected the "prev_version" of 'r2', 'r6', 'r9' to point to 'r0', 'r4', 'r7', respectively:
```
In [3]: with h5py.File(Filename(d) / 'data.h5', 'w') as f:
   ...:         sv['foo'] = np.array([1, 2, 3])
   ...:
[PYFLYBY] from deshaw.file import Filename
[PYFLYBY] import h5py
[PYFLYBY] import numpy as np
[PYFLYBY] from versioned_hdf5 import VersionedHDF5File

In [4]: with h5py.File(Filename(d) / 'data.h5', 'r+') as f:
   ...:     for i in range(1, 11):
   ...:         with vf.stage_version(f'r{i}') as sv:
   ...:             sv['foo'][:] = np.array([1, i, 3])
   ...:

In [5]: with h5py.File(Filename(d) / 'data.h5', 'r') as f:
   ...:     print(f['_version_data/versions'].keys())
   ...:
<KeysViewHDF5 ['__first_version__', 'r0', 'r1', 'r10', 'r2', 'r3', 'r4', 'r5', 'r6', 'r7', 'r8', 'r9']>

In [6]: from versioned_hdf5.replay import delete_versions
   ...: with h5py.File(Filename(d) / 'data.h5', 'r+') as f:
   ...:     delete_versions(f, ['r1', 'r5', 'r8'])
   ...:

In [11]: with h5py.File(Filename(d) / 'data.h5', 'r') as f:
    ...:     for v in f['_version_data/versions']:
    ...:         print(f"{v} -> {f['_version_data/versions'][v].attrs.get('prev_version', None)}")
    ...:
__first_version__ -> None
r0 -> __first_version__
r10 -> r9
r2 -> r1
r3 -> r2
r4 -> r3
r6 -> r5
r7 -> r6
r9 -> r8
```
As you can see the "prev_version" is still pointing to the old version (prior to the `delete_versions` call).

asmeurer:
Fixes #240.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,12 @@
 Versioned HDF5 Change Log
 =========================
 
+## 1.3.13 (2023-01-30)
+
+### Minor Changes
+
+- Fix the setting of `prev_version` in `delete_versions()`.
+
 ## 1.3.12 (2022-08-17)
 
 ### Minor Changes
--- a/rever.xsh
+++ b/rever.xsh
@@ -7,7 +7,7 @@ $PROJECT = 'versioned-hdf5'
 def run_tests():
     # Don't use the built-in pytest action because that uses Docker, which is
     # overkill and requires installing Docker
-    with run_in_conda_env(['python=3.8', 'pytest', 'numpy', 'h5py<3',
+    with run_in_conda_env(['python=3.8', 'pytest', 'numpy', 'h5py',
                            'ndindex', 'pyflakes', 'pytest-cov',
                            'scipy', 'pytest-doctestplus', 'pytest-flakes',
                            'doctr', 'sphinx']):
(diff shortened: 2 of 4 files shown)


asmeurer:
Looks like there are some test failures that are likely caused by upstream packages. It's not clear to me that any of these are bugs in versioned-hdf5. I'll make sure the tests pass locally at least before merging.

asmeurer:
I opened issues for them. We'll need to look into them at some point. Let me know if you want me to prioritize them.

asmeurer:
Released 1.3.13 with these changes.

PR #247, opened 2023-05-23
Fix InMemoryDataset.__getitem__ to work with development version of h5py

### LINKED ISSUE #246: `InMemoryDataset.__getitem__` is broken if `new_dtype` is not specified with h5py 3.8
In trying to debug #244, I found I was unable to examine the contents of a file in the following way with the current development version of `h5py`:

```python
import shutil
import tempfile

import h5py
import numpy as np
import subprocess
from versioned_hdf5 import VersionedHDF5File, delete_versions


if __name__ == '__main__':
    d = tempfile.mkdtemp()
    fn = d + '/data.h5'
    try:

        # Write a file with no data
        with h5py.File(fn, 'w') as f:
            vf = VersionedHDF5File(f)
                sv.create_dataset('key0', data=np.array([1, 2, 3]), maxshape=(None,),
                                  chunks=(10000,), compression='lzf')
                print(sv['key0'][:])

        # Create a new Dataset, resizing in place, saving a new version
        with h5py.File(fn, 'r+') as f:
            vf = VersionedHDF5File(f)
                print(sv['key0'][:])

    finally:
        shutil.rmtree(d)
```

It seems likely that using the private API [here](https://github.com/deshaw/versioned-hdf5/blob/e7cc0f8c06fdb93af0a729e387898daa979b3a42/versioned_hdf5/wrappers.py#L599) broke at some point in the past.

peytondmurray:
This PR fixes versioned-hdf5 to work with the current development version of `h5py`, and removes an instance where we were making an `h5py` internal API call. To ensure we don't break compatibility with future `h5py` versions, I've also added a CI job to run against the development version.

Additionally this PR aligns itself with [NEP29 for python version support](https://numpy.org/neps/nep-0029-deprecation_policy.html), which simplifies the CI.

Closes #246.

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -595,9 +595,6 @@ class InMemoryDataset(Dataset):
         # This boilerplate code is based on h5py.Dataset.__getitem__
         args = args if isinstance(args, tuple) else (args,)
 
-        if new_dtype is None:
-            new_dtype = getattr(self._local, 'astype', None)
-
         # Sort field names from the rest of the args.
         names = tuple(x for x in args if isinstance(x, str))
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,18 +5,33 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
-        h5py-version: ['3.6', '3']
+        python-version: ['3.9', '3.10', '3.11']
+        h5py-version: ['3.8', 'dev']
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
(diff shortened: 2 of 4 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -595,9 +595,6 @@ def __getitem__(self, args, new_dtype=None):
         # This boilerplate code is based on h5py.Dataset.__getitem__
         args = args if isinstance(args, tuple) else (args,)
 
-        if new_dtype is None:
-            new_dtype = getattr(self._local, 'astype', None)
-
```
ArvidJB, line 600: Looks like there was a change in h5py which removed the same code you are removing:
https://github.com/h5py/h5py/commit/dcc694b2809c94cfa4f9718eedba5d61ab55318a
According to the comment above

> This boilerplate code is based on h5py.Dataset.__getitem__

this looks good.

ArvidJB (approved):

Issue #244, opened 2023-02-18
Error when combining delete_versions and h5repack on empty Dataset


ArvidJB:
We are encountering a strange error when we use `delete_versions` on an empty `Dataset` and run `h5repack`. Here's the smallest reproducer I could come up with:
```
import shutil
import tempfile

import h5py
import numpy as np
import subprocess
from   versioned_hdf5           import VersionedHDF5File, delete_versions

if __name__ == '__main__':
    d = tempfile.mkdtemp()
    fn = d + '/data.h5'
    try:
        with h5py.File(fn, 'w') as f:
            vf = VersionedHDF5File(f)
            with vf.stage_version('r0') as sv:
                sv.create_dataset('key0', data=np.array([]), maxshape=(None,),
                                  chunks=(10000,), compression='lzf')

        with h5py.File(fn, 'r+') as f:
            vf = VersionedHDF5File(f)
            with vf.stage_version('r1') as sv:
                sv['key0'].resize((0,))

        tfn = tempfile.mktemp(dir=d)
        shutil.copy(fn, tfn)
        with h5py.File(tfn, 'r+') as f:
            delete_versions(f, ['r0'])
        p = subprocess.Popen(['h5repack', tfn, fn])
        p.wait()

        with h5py.File(fn, 'r+') as f:
            vf = VersionedHDF5File(f)
            with vf.stage_version('r2') as sv:
                sv['key0'].resize((0,))

        with h5py.File(fn, 'r') as f:
            vf = VersionedHDF5File(f)
            cv = vf[vf.current_version]
            print(cv['key0'][:])
    finally:
        shutil.rmtree(d)
```
This gives the following error
```
Traceback (most recent call last):
  File "/u/bessen/opensource/versioned-hdf5/delete_versions_h5repack_bug_5.py", line 49, in <module>
    print(cv['key0'][:])
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/codemill/bessen/versioned_hdf5_env/lib64/python3.10/site-packages/h5py/_hl/dataset.py", line 768, in __getitem__
    return self._fast_reader.read(args)
  File "h5py/_selector.pyx", line 376, in h5py._selector.Reader.read
OSError: Can't read data (src and dest dataspaces have different number of elements selected)

Process finished with exit code 1
```
[...]

Issue #239, opened 2022-09-27
Use hypothesis in the tests


asmeurer:
Right now the tests in the test suite all use hard coded examples to try to hit different code paths. But the tests could be a lot more robust if they made use of [hypothesis](https://hypothesis.readthedocs.io/en/latest/). 

An easy place to start here would be the (currently very slow) reshape tests, which use a hard coded parameterization of different values, but could easily just use hypothesis strategies. 

More generally, we would need to create some hypothesis strategies to represent the process of creating versions/datasets, perhaps by making use of [state machine](https://hypothesis.readthedocs.io/en/latest/stateful.html) strategies.

PR #248, opened 2023-06-01
Correctly handle deletion of empty datasets

peytondmurray:
This PR modifies how empty datasets are deleted, closing #244.

## Changes

* As I was debugging, I broke out function definitions from inside `delete_versions` and `_recreate_raw_data` to help separate logic a little  more. I don't think we gain anything by redefining `_walk`, `_get_np_fillvalue`, and `_delete_dataset` during each call, but I can change this back if it's just confusing.
* Added a bunch of type annotations for the functions I touched, it really helped with debugging once I knew what types different objects were.
* When the user deletes one or more versions, and those versions contain one or more datasets which don't exist in the versions being kept (and therefore the raw data needs to be deleted), we now skip recreating the raw data, recalculating hash tables, and recreating the virtual datasets associated with the dataset.
This is done because empty datasets cannot be virtual; see https://github.com/deshaw/versioned-hdf5/commit/85507e91576e197f50d43f50d0abe66e453c32cb. In other words, empty datasets at `_version_data/versions/<version>/<name>` are real, and since they don't refer to data in `_version_data/<name>/raw_data` as they would if they held actual data, they can safely just be deleted. As far as I can tell, this is a limitation of `h5py` and `hdf5` being unable to store empty virtual datasets.
* Added a test to cover the reproducer from #244.

### DIFF
--- a/versioned_hdf5/replay.py
+++ b/versioned_hdf5/replay.py
@@ -1,9 +1,20 @@
-from h5py import VirtualLayout, h5s, Dataset, Group, __version__ as h5py_version
+from __future__ import annotations
+from typing import List, Iterable, Union, Dict, Any
+from h5py import (
+    VirtualLayout,
+    h5s,
+    HLObject,
+    Dataset,
+    Group,
+    File,
+    __version__ as h5py_version
+)
 from h5py._hl.vds import VDSmap
 from h5py._hl.selections import select
 from h5py.h5i import get_name
 
 from ndindex import Slice, ChunkSize, Tuple
+from ndindex.ndindex import NDIndex
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


versioned_hdf5/replay.py
```diff
@@ -316,9 +347,143 @@ def _recreate_virtual_dataset(f, name, versions, raw_data_chunks_map, tmp=False)
...
+    versions = version_data['versions']
+
+    if name == 'versions':
+        return
+
+    versions_to_keep = set(versions) - set(versions_to_delete)
+
+    # If the dataset is empty in the versions to delete, we don't
+    # need to recreate the raw data, hash table, or virtual datasets.
+    if _all_extant_are_empty(f, name, versions_to_delete):
```
asmeurer, line 434: Is this is the only effective change here?

versioned_hdf5/replay.py
```diff
@@ -117,7 +127,43 @@ def tmp_group(f):
         tmp = f['_version_data/__tmp__']
     return tmp
 
-def _recreate_raw_data(f, name, versions_to_delete, tmp=False):
+# See InMemoryDataset.fillvalue. In h5py3 variable length strings use None
+# for the h5py fillvalue, but require a string fillvalue for NumPy.
+def _get_np_fillvalue(data: Dataset) -> Union[bytes, float]:
```
asmeurer, line 132: This is not strictly correct (the fillvalue can be any NumPy scalar type supported by h5py), but it doesn't really matter because type annotations aren't actually used anywhere.

versioned_hdf5/replay.py
peytondmurray: Yep.

versioned_hdf5/replay.py
```diff
@@ -1,9 +1,19 @@
-from h5py import VirtualLayout, h5s, Dataset, Group, __version__ as h5py_version
+from typing import List, Iterable, Union, Dict
```
asmeurer, line 1: Probably doesn't matter for what you've added, but let's use `from __future__ import annotations` and `if TYPE_CHECKING` to prevent type annotations from having any runtime effects.

versioned_hdf5/replay.py
peytondmurray: Hmm, maybe I'll just change this to `Any` for now.

asmeurer:
It's been a while, but it seems correct. If you want to be extra sure you could try testing more extensive combinations of different versions including or not including the dataset with and without non-fill data. 

I do rather wish I had made use of hypothesis when developing this library (https://github.com/deshaw/versioned-hdf5/issues/239), as its state machine strategies would be a perfect fit for extensively testing this sort of thing. It's something I would heavily suggest adding if we ever decide to add any significant new functionality.

versioned_hdf5/replay.py
peytondmurray: I'll add `from __future__ import annotations`, but I'm not sure how to make use of `if TYPE_CHECKING`. I'm not actually using any types defined in `versioned_hdf5` for static type checks except `VersionedHDF5File`, which is required to be defined before this module is executed anyway because it's used for some runtime type checks elsewhere in this file. If there's a better way to handle this though I'd be grateful to learn.

peytondmurray:
I really like the idea of making hypothesis do the heavy lifting for the tests. I'm going to try to spend a little more time doing some general upkeep on this library, so there's a good chance I'll do this in the future.

versioned_hdf5/replay.py
asmeurer: Oh I guess you're right that `if TYPE_CHECKING` is only useful if you have types other than from `typing`.

ArvidJB (approved):
I defer to @asmeurer, looks good to me, though!

Your observation that empty datasets are "special" and cannot be virtual seems to explain what's going on here. I think whenever we ran into this problem it did involve versions which were empty datasets.

PR #251, opened 2023-06-27
Add info log showing stats about data upon exiting stage_version context

### LINKED ISSUE #166: Log statistics at exit of stage_version context
I think it would be helpful if we could log some statistics after writing each version. So probably at the end of a `stage_version` context?
I would be interested in some information like
```
dataset_name: old_shape -> new_shape, chunk size, number of new raw chunks written, number of chunks reused
```
for every dataset impacted in the transaction.

peytondmurray:
This PR logs a message when exiting the `VersionedHDF5File.stage_version` context manager displaying some useful information about the data written. Information is displayed at the `INFO` log level.

This PR closes #166.

Here's a simple example taken from a test:

```python
import logging
import pathlib

import h5py
import numpy as np

logging.basicConfig(level=logging.INFO)

dt = np.dtype('double')

# Clean out old data files
for p in pathlib.Path().glob("*.h5"):
    p.unlink()

# Create new versioned file
    vf = VersionedHDF5File(f)
        sv.create_dataset('bar', (2, 15220, 2),
                          chunks=(300, 100, 2),
                          dtype=dt, data=np.full((2, 15220, 2), 0, dtype=dt))
        sv.create_dataset('baz', (1, 10, 2),
                          chunks=(600, 2, 4),
                          dtype=dt, data=np.full((1, 10, 2), 0, dtype=dt))

    vf = VersionedHDF5File(f)
        bar = sv['bar']
        bar.resize((3, 15222, 2))
        baz = sv['baz']
        baz.resize((1, (4) * 10, 2))
        baz[:, -10:, :] = np.full((1, 10, 2), 3, dtype=dt)
```

Running this produces displays the following:

```
INFO:versioned_hdf5.api:
  bar: Shape: None -> (2, 15220, 2); Chunks: None -> (300, 100, 2); New chunks written: 2; Number of chunks reused: 0
  baz: Shape: None -> (1, 10, 2); Chunks: None -> (600, 2, 4); New chunks written: 1; Number of chunks reused: 0
INFO:versioned_hdf5.api:
  bar: Shape: (2, 15220, 2) -> (3, 15222, 2); Chunks: ChunkSize((300, 100, 2)) -> ChunkSize((300, 100, 2)); New chunks written: 0; Number of chunks reused: 4
  baz: Shape: (1, 10, 2) -> (1, 40, 2); Chunks: ChunkSize((600, 2, 4)) -> ChunkSize((600, 2, 4)); New chunks written: 0; Number of chunks reused: 2
```

@ArvidJB @quarl What do you think of this? As long as this looks good to you, I'll move this out of draft and mark it ready for review.

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -4,7 +4,9 @@ Public API functions
 Everything outside of this file is considered internal API and is subject to
 change.
 """
+import logging
 import numpy as np
+from typing import Set, Optional
 
 from contextlib import contextmanager
 import datetime
@@ -14,6 +16,10 @@ from .versions import (create_version_group, commit_version,
                        get_version_by_timestamp, get_nth_previous_version,
                        set_current_version, all_versions, delete_version, )
 from .wrappers import InMemoryGroup
+from .hashtable import Hashtable
+
+
+logger = logging.getLogger(__name__)
 
 
 class VersionedHDF5File:
@@ -195,6 +201,7 @@ class VersionedHDF5File:
         group = create_version_group(self.f, version_name,
                                      prev_version=prev_version)
 
+
         try:
             yield group
             group.close()
@@ -203,6 +210,9 @@ class VersionedHDF5File:
                            compression=group.compression,
                            compression_opts=group.compression_opts,
                            timestamp=timestamp)
+
+            self._log_version_diff_stats(old_current, self.current_version)
+
         except:
             delete_version(self.f, version_name, old_current)
             raise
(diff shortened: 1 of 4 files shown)


versioned_hdf5/api.py
```diff
@@ -230,3 +240,69 @@ def __repr__(self):
         else:
             return f"<VersionedHDF5File object \"{self.f.filename}\" (mode" \
                    f" {self.f.mode})>"
+
+    def get_hashes(self, name: str) -> Set[bytes]:
```
ArvidJB, line 244: Does this need to be public?

versioned_hdf5/api.py
```diff
@@ -230,3 +240,69 @@ def __repr__(self):
...
+            chunks_reused = len(old_hashes & new_hashes)
+            new_chunks_written = len(new_hashes - old_hashes)
+
+            msg.append(
+                f"  {name}: Shape: {old_shape} -> {new_shape}; "
+                f"Chunks: {old_chunks} -> {new_chunks}; "
+                f"New chunks written: {new_chunks_written}; "
+                f"Number of chunks reused: {chunks_reused}"
+            )
+        logger.info("\n".join(msg))
```
ArvidJB, line 308: This should log at debug level by default, I think. Most of our production jobs run with info level debugging and would end up printing this, which our end users probably do not care about.

ArvidJB:
Looks good to me! Just make sure the logging is at debug level, not info.

versioned_hdf5/api.py
peytondmurray: Definitely not, I'll change it now.

versioned_hdf5/api.py
peytondmurray: Sounds good to me.

peytondmurray:
@asmeurer Before merging this I thought I'd ask you about the implementation of `__bool__` here, since it is user facing. Previously doing `bool(dataset)` would fail if the dataset was a scalar array because it would fall back to checking `len(dataset)`. With this change, scalar datasets will no longer throw errors. Thoughts on this behavior?

asmeurer:
As I mentioned on Slack, the behavior should mirror h5py, which itself tries to mirror numpy. `bool(0_d_array)` works in numpy, and I haven't checked but it ought to work in h5py, so this change sounds correct (len(0_d) should fail, but I think len is just used as the default implementation of `bool`).

PR #257, opened 2023-07-13
Fix object dtype hashing

### LINKED ISSUE #256: Chunks incorrectly reused with object dtype datasets
Adding a dataset with an object dtype (e.g., `h5py.string_dtype(length=None)`) can result in incorrect data being stored due to incorrect chunk reuse over multiple versions. Code to reproduce the issue:

```python
import h5py
import versioned_hdf5
import numpy as np

N = 100

with h5py.File("test.h5", mode="w") as f:
    file = versioned_hdf5.VersionedHDF5File(f)

    s = ""
    for i in range(N):
        s += "a"
        arr = np.array([s], dtype=object)

        with file.stage_version(f"r{i}") as group:
            group.create_dataset(
                "values", shape=(1,), dtype=h5py.string_dtype(length=None), data=arr
            )

with h5py.File("test.h5", mode="r") as f:
    file = versioned_hdf5.VersionedHDF5File(f)
    for i in range(N):
        print(file[f"r{i}"]["values"][()])
```

For me, this outputs:
```
[b'a']
[b'aa']
[b'aaa']
[b'aaaa']
[b'aaaaa']
[b'aaaaaa']
[b'aaaaaaa']
[b'aaaaaaaa']
[b'aaaaaaaaa']
[b'aaaaaaaaaa']
[b'aaaaaaaaaaa']
[b'aaaaaaaaaaaa']
[b'aaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
[b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa']
```

As I understand it, this is because object addresses can be reused, even for strings.

https://github.com/deshaw/versioned-hdf5/blob/b65e54cc22c81114c6f3b45fcf6bc9f9cffb91b7/versioned_hdf5/hashtable.py#L77-L78

`Hashtable.hash` computes a hash on `data.data.tobytes()`, which in the case of a `dtype=object` array is an array of bytes representing the addresses of each object. This issue no longer occurs if we generate a random hash instead.

Thanks!

peytondmurray:
Calling `arr.tobytes()` on an object array returns the object address in bytes format, not the byte-representation of the underlying data. This makes it unsuitable for hashing. In this PR a branch is introduced to handle object dtypes specially, hashing each element directly, and returning the result. Handling of other dtypes remains unchanged. A new test was added to catch this in the future (although due to the fact that object references may or may not be reused, this test isn't always guaranteed to fail if this issue comes up again, although it seems reasonably reliable.)

Closes #256.


asmeurer:
Might be worth thinking about what happens to datasets that were created before this change since this changes hash values.

peytondmurray:
:point_up: Very good point, something I overlooked before. I can think of a couple of ways to mitigate the issue:

1. Use a flag or environment variable to force use of old hashing method for object dtypes
2. Look at the `timestamp` attribute for the version to determine which hashing method to use. If it's before a certain date, use the old method. I don't really like this idea as much as the first option, because the behavior will be very unintuitive if it depends on the system clock :/

Do you have any preferences here?

ArvidJB:
I don't think we have to worry too much about backwards-compatibility. If I understand it correctly all previous hashes for dtype='object' arrays were basically random numbers and there was no reuse happening?

What happens if arbitrary bytes are in the array? For example we sometimes store pickle byte strings in HDF5 Datasets as well:
```
In [1]: a = np.array([['foo', {'bar': 1.23}], ('hello', 'world', 3.14)], dtype='O')
[PYFLYBY] import numpy as np

In [2]: b = np.array([pickle.dumps(x) for x in a], dtype='O')
[PYFLYBY] import pickle

In [3]: b
Out[3]:
array([b'\x80\x04\x95\x1d\x00\x00\x00\x00\x00\x00\x00]\x94(\x8c\x03foo\x94}\x94\x8c\x03bar\x94G?\xf3\xae\x14z\xe1G\xaese.',
       b'\x80\x04\x95\x1c\x00\x00\x00\x00\x00\x00\x00\x8c\x05hello\x94\x8c\x05world\x94G@\t\x1e\xb8Q\xeb\x85\x1f\x87\x94.'],
      dtype=object)
```
Can you add a test for this? Not sure if the `bytes(str(...))` conversion is correct in that case?

asmeurer:
What actually is the `bytes` for object arrays? Is it possible for another non-object array to have the same bytes? If so, there will be a hash collision.

asmeurer:
If a new version of versioned-hdf5 gives different hashes than before it could break things. I'm not sure. It's something that would need to be tested. 

It might also be a good idea to add a file version string somewhere to the file, so that it would be possible to have a "version 2" of versioned-hdf5 files so that files generated by older versions could be updated.

peytondmurray:
> What actually is the bytes for object arrays?

Object dtype arrays store the addresses of their elements. When you call `arr.tobytes()`, what you get is the concatenation of the object ids as bytes:

image

> Is it possible for another non-object array to have the same bytes? If so, there will be a hash collision.

Definitely:

image

Here I've made an int64 dtype array containing the ids of the object array from above. It's got the same byte representation.

---

I've added an attr for a data version identifier in the previous commit. A warning is issued to the user if `versioned-hdf5` is expecting a different data version identifier than is found in the file being read. This should at least partially mitigate the issues discussed before.

How worried should we be about hash collisions here? It seems like a pretty contrived example, though not impossible. If we want to include other information (e.g. dtype) to the hash to prevent object dtype arrays from colliding with non-object dtype arrays that could work too.

ArvidJB:
To check against collisions can you add an array comparison check in `write_dataset`? If I understand correctly that's where we use the hash table to look up the slice for `data_s`. We should check the array slices are the same, something like
```
                data_hash = hashtable.hash(data_s)
                raw_slice2 = hashtable.setdefault(data_hash, raw_slice)
                if raw_slice2 == raw_slice:
                    slices_to_write[raw_slice] = s
                else:
                    # reuse an existing slice, check data to guard against hash collisions
                    assert np.all(ds[raw_slice] == data_s)
```

It's extremely unlikely, but we should be aware that for older files where the hashes are based on the memory address and we could end up reusing a chunk with different data.

versioned_hdf5/api.py
```diff
@@ -67,6 +68,17 @@ def __init__(self, f):
         self.f = f
         if '_version_data' not in f:
             initialize(f)
+        else:
+            # This is not a new file; check data version identifier for compatibility
+            if self.data_version_identifier != DATA_VERSION:
+                warnings.warn(
```
ArvidJB, line 74: I think that's the right idea, but unfortunately we have 100,000s of old files which would then trigger those warnings. Can we make them suppressed by default?

versioned_hdf5/hashtable.py
```diff
@@ -75,7 +75,18 @@ def from_raw_data(cls, f, name, chunk_size=None, hash_table_name='hash_table'):
...
 
     def hash(self, data):
-        return self.hash_function(data.data.tobytes() + bytes(str(data.shape), 'ascii')).digest()
+        # Object dtype arrays store the ids of the elements, which may or may not be
+        # reused, making it unsuitable for hashing. Instead, we need to make a combined
+        # hash with the value of each element.
+        if data.dtype == 'object':
+            hash_value = self.hash_function()
+            for value in data.flat:
+                hash_value.update(bytes(str(value), 'utf-8'))
```
ArvidJB, line 84: We should probably also add `data.shape` like below?

versioned_hdf5/api.py
peytondmurray: Would a log message be better? Maybe emitted at `INFO` level, so it's suppressed by default?

versioned_hdf5/hashtable.py
peytondmurray: Good idea.

versioned_hdf5/api.py
peytondmurray: I made this an INFO level log message here.

peytondmurray:
Okay, updated this branch:

- Added `data.shape` to the hash
- Changed the warning to a log message at `INFO` level

There's still an issue here. After adding @ArvidJB's suggested element-by-element check when you try to write data which has a chunk with the same hash as a chunk that already exists in the file, I find that nearly all the tests fail. 

Running `pytest --pdb -svv versioned_hdf5/tests/test_api.py::test_stage_version` fails to write the data because writing data happens in two steps:

1. The hash table is compared with the slices to be written.
  a. If the slice is already in the hash table, don't bother writing the data again
  b. Otherwise, record the data hash in the hash table, and add it to the list of data to write
2. The slices which need to be written to the file are actually written

So the issue is that the hash table gets updated _before_ data actually gets written to the file. I think this part needs to be refactored so that doing lookups in the hash table happens in the same iteration as the data writing.

@ArvidJB In short I think I need a little bit more time to complete this - FYI

ArvidJB:
Does that mean we could currently potentially populate the hash table and then potentially not write the data to the file (process is aborted, crashes, ...). Ideally we'd do these two things in a "transaction"?

peytondmurray:
Possibly? There's a try/except block in `stage_version` to prevent getting partway through writing a version, encountering an error, and then leaving garbage in the file: https://github.com/deshaw/versioned-hdf5/blob/a6d57df8527c7febb23b44685b2567b08c1b8665/versioned_hdf5/api.py#L205.

But if you look at [`delete_version`](https://github.com/deshaw/versioned-hdf5/blob/a6d57df8527c7febb23b44685b2567b08c1b8665/versioned_hdf5/versions.py#L160) it doesn't look like the hash table gets cleaned up here. I'm wondering whether when `delete_versions` is called we need to iterate through every chunk and remove any hashes that don't point to real data anymore - using those slices to access data could cause an issue. @asmeurer What are your thoughts on this?

asmeurer:
I would be careful about the performance implications of doing an equality check guard for hash collisions. 

> Here I've made an int64 dtype array containing the ids of the object array from above. It's got the same byte representation.

Presumably that couldn't happen though because it's a different dtype. But if all it does is hash the id then I suppose any dtype could in principle have a collision, although maybe it's unlikely enough to not worry about here. The biggest question is handling files that were already written with the old hash values.

peytondmurray:
> I would be careful about the performance implications of doing an equality check guard for hash collisions.

Yep, we'd be giving up a lot of speed for this - any time you write a chunk that already exists in the file you'd be doing element-by-element comparison. Almost feels like it defeats the purpose of the hash table. As far as I can tell, it's the only way to guard absolutely against hash collisions. I'm happy to take it it out if that makes sense - or to only do this for object dtypes.

> Presumably that couldn't happen though because it's a different dtype. 

Good point, dtypes of datasets can't be changed, right? Then this example would never happen in practice.

> But if all it does is hash the id then I suppose any dtype could in principle have a collision, although maybe it's unlikely enough to not worry about here. The biggest question is handling files that were already written with the old hash values.

No, only numpy arrays where the elements of the array _are_ ids would be affected. To my knowledge this is only the object dtype for now, although the proposed variable-length string dtype would also work the same way.

peytondmurray:
After further discussion with @asmeurer, it was decided that an element-by-element check each time a chunk was reused would result in an unacceptable performance penalty. Instead, I've implemented the following:

1. Upon opening a file written with an outdated data version using file mode `r+`, we rebuild all the hash tables. Other modes of reading/writing to the file are unchanged.
2. To rebuild the hash table, we first delete the existing one. Then we iterate through all versions of the dataset, updating the hash table as we go, emulating the how data gets written to the file in the first place (but this time we only write to the hash table, we don't actually write any new raw data).
3. Finally, we update the data version of the file to the most recent version.

I added some tests for this - both to make sure we are writing new, stable hashes on object dtype arrays, and to make sure that the hash tables rebuilt for other dtypes remain unchanged. So far it all looks good.

versioned_hdf5/api.py
```diff
@@ -67,6 +67,26 @@ def __init__(self, f):
         self.f = f
         if '_version_data' not in f:
             initialize(f)
+        else:
+            # This is not a new file; check data version identifier for compatibility
+            if self.data_version_identifier != DATA_VERSION:
```
asmeurer, line 72: Should probably be a < check here. Actually there should maybe be some sort of error in the VersionedHDF5File constructor if the data version is greater than expected.

versioned_hdf5/backend.py
```diff
@@ -5,6 +5,7 @@
 from .hashtable import Hashtable
 
 DEFAULT_CHUNK_SIZE = 2**12
+DATA_VERSION = 'v2'
```
asmeurer, line 8: Let's use an integer instead of a string, to make comparisons easier.

asmeurer:
I can't tell what's in the binary files here, but it's a good idea to include binary files for testing files with old data versions. We'll probably want to split them out from the library code so they don't get shipped with the release.

versioned_hdf5/api.py
```diff
@@ -91,6 +111,23 @@ def current_version(self):
         """
         return self._versions.attrs['current_version']
 
+    @property
```
asmeurer, line 114: We can create a setter property here too.

versioned_hdf5/api.py
peytondmurray: Good idea, I've changed this to raise an exception if a user loads a file written by a future version as well.

versioned_hdf5/api.py
peytondmurray: :+1:

versioned_hdf5/backend.py
peytondmurray: I like it. I considered moving to `packaging.version.Version`, but I think this is just simpler.

peytondmurray:
> I can't tell what's in the binary files here, but it's a good idea to include binary files for testing files with old data versions. We'll probably want to split them out from the library code so they don't get shipped with the release.

Yep, good idea. I've moved the test data into a top-level `test_data` directory, which shouldn't get shipped with the package. It looks like we're still shipping `versioned_hdf5/tests` through, but maybe we can tackle that in a different issue.

You can see the contents of the file by looking at the `generate_bad_data` function, but in summary:

* `object_dtype_bad_hashtable_data.h5` contains the following array in only one version:
```python
arr = np.array(
    ["abcd", "def", "ghi", "jkl"],
    dtype=object
)
```
* `object_dtype_bad_hashtable_data2.h5` contains a few arrays in multiple versions. First, we write to version `r0`:
```python
arr = np.array(
    ["abcd", "def", "ghi", "jkl"],
    dtype=object
)
arr2 = np.array(
    ["abcd", "pqrs", "df", "tuvw", "xyz"],
    dtype=object
)
arr3 = np.linspace(1, 100, 1000)
```

Then we write to version `r1`:
```python
arr1[1] = "foo"
arr3[3] = 8
```

Then in version `r2` we write:
```python
arr1[2] = "bar"
arr2[2] = "what"
```

This way there are multiple versions of each dataset with multiple dtypes tested.

peytondmurray:
@asmeurer Let me know if this looks good to you, then we can merge it. We should also probably publish a release for this (and test the new release machinery as well).

versioned_hdf5/backend.py
asmeurer: File versions are usually just simple integers. There's no need to be semantic here.

asmeurer:
Let's make sure the file version number is tested. It's possible I missed the test for it. Everything else looks fine to me.

peytondmurray:
There's a test [here](https://github.com/peytondmurray/versioned-hdf5/blob/da373bacc2bda4283ee734d5747b629a3b88d437/versioned_hdf5/tests/test_api.py#L1886) which checks whether a new file written correctly writes the expected `DATA_VERSION`, and another [here](https://github.com/peytondmurray/versioned-hdf5/blob/da373bacc2bda4283ee734d5747b629a3b88d437/versioned_hdf5/tests/test_api.py#L1901) which checks that files without a valid data version identifier issue a log message. Is there another I could add here? :thinking:

Issue #258, opened 2023-07-18
Move away from `rever` for releases


peytondmurray:
`rever` is not being actively developed anymore. I think it should be possible to cover most if not all of what `rever` currently does using github actions.

peytondmurray:
Here's what we'll need to implement to replace `rever.xsh`:

- [ ] run tests
- [ ] build docs
- [ ] annotated tag
- [ ] push tag
- [ ] pypi publish
- [ ] publish gh-pages

Some of these are already complete.

ArvidJB:
Is this a prerequisite for #253 or can we still cut a release using rever?

peytondmurray:
Asking @asmeurer about this now.

peytondmurray:
Release https://github.com/deshaw/versioned-hdf5/releases/tag/1.3.14 has been made. Unless plans change, the next release will be made without `rever`.

peytondmurray:
I updated PYPI to add the `pypi_publish.yml` workflow as a trusted publisher. Closing this now as complete.

PR #260, opened 2023-08-15
Add workflow to publish to pypi upon release tag

peytondmurray:
This PR adds a workflow to publish the package to PYPI using [trusted publishing.](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) We'll need to add this workflow as a trusted publisher for the project before the next release.

Partially addresses #258.

### DIFF
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -1,0 +1,67 @@
+name: Release Python Bindings to PyPI
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


asmeurer:
This looks fine, but as with https://github.com/deshaw/versioned-hdf5/pull/255, the only way to really know is to test it. I guess we can somehow use test pypi to do a test release. Otherwise it might be a good idea to just to a basic 1.3.14.1 release with no changes in it after this is merged just to check that everything works correctly. If it breaks, we can yank the release and fix whatever the problem is.

PR #262, opened 2023-08-25
Avoid quadratic runtime in delete_version

### LINKED ISSUE #261: delete_versions is very slow
Using `delete_versions` is surprisingly slow. I created a small script which reproduces the slow performance:
```
import logging
import shutil
import tempfile

import h5py
import numpy as np
from   versioned_hdf5           import VersionedHDF5File, delete_versions

if __name__ == '__main__':
    logging.basicConfig(
        format="[%(asctime)s:%(msecs)03d %(levelname)s] %(message)s",
        level=logging.INFO,
        datefmt='%Y-%m-%d %H:%M:%S')
    d = tempfile.mkdtemp()
    fn = d + '/data.h5'
    try:
        logging.info('first write')
        with h5py.File(fn, 'w') as f:
            vf = VersionedHDF5File(f)
                sv.create_dataset('values', data=np.zeros((1, 100)), fillvalue=0,
                                  chunks=(300, 120), maxshape=(None, None), compression='lzf')
        logging.info('updates')
        with h5py.File(fn, 'r+') as f:
            vf = VersionedHDF5File(f)
            for i in range(1, 101):
                for j in range(max(0, i - 3), i + 1):
                    cs = np.arange(10)
                    np.random.shuffle(cs)
                    for c in cs:
                        with vf.stage_version(f'r{i}_{j}_{c}') as sv:
                            if j >= sv['values'].shape[0]:
                                sv['values'].resize((j + 1, 100))
                            sv['values'][j, c * 10:(c + 1) * 10] = j * 10 + c

        logging.info('keep only every 10th version')
        versions_to_delete = []
        with h5py.File(fn, 'r') as f:
            vf = VersionedHDF5File(f)
                if i % 10 != 0:

        logging.info('delete_versions')
        with h5py.File(fn, 'r+') as f:
            delete_versions(f, versions_to_delete)

        logging.info('finished')
    finally:
        shutil.rmtree(d)
```
Running this prints
```
$ time python version_retention_speed_1.py
[2023-08-25 12:53:00:695 INFO] first write
[2023-08-25 12:53:00:704 INFO] updates
[2023-08-25 12:53:40:298 INFO] keep only every 10th version
[2023-08-25 12:53:40:551 INFO] delete_versions
[2023-08-25 12:59:40:088 INFO] finished

real    6m39.668s
user    6m37.740s
sys     0m0.421s
```
Most of the time is taken by the logic to fix up `prev_version`:
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   483
   484                                           def delete_versions(
   485                                               f: Union[VersionedHDF5File, File],
   486                                               versions_to_delete: Iterable[str]
   487                                           ):
   488                                               """Completely delete the given versions from a file
   489
   490                                               This function should be used instead of deleting the version group
   491                                               directly, as this will not delete the underlying data that is unique to
   492                                               the version.
   493                                               """
   494         1          5.0      5.0      0.0      if isinstance(f, VersionedHDF5File):
   495                                                   f = f.f
   496
   497         1         58.0     58.0      0.0      version_data = f['_version_data']
   498         1          1.0      1.0      0.0      if isinstance(versions_to_delete, str):
   499                                                   versions_to_delete = [versions_to_delete]
   500
   501         1         38.0     38.0      0.0      versions = version_data['versions']
   502
   503         1         95.0     95.0      0.0      if '__first_version__' in versions_to_delete:
   504                                                   raise ValueError("Cannot delete first version")
   505
   506      3575       1740.0      0.5      0.0      for version in versions_to_delete:
   507      3574      42918.0     12.0      0.0          if version not in versions:
   508                                                       raise ValueError(f"Version {version!r} does not exist")
   509
   510         1        152.0    152.0      0.0      current_version = versions.attrs['current_version']
   511         2        104.0     52.0      0.0      while current_version in versions_to_delete:
   512         1        129.0    129.0      0.0          current_version = versions[current_version].attrs['prev_version']
   513
   514         2    1139821.0 569910.5      0.2      for name in _walk(version_data):
   515         1    1292915.0 1292915.0      0.2          _delete_dataset(f, name, versions_to_delete)
   516
   517      3575       3773.0      1.1      0.0      for version_name in versions_to_delete:
   518      3574     348503.0     97.5      0.1          prev_version  = versions[version_name].attrs['prev_version']
   519   7814551   19292715.0      2.5      3.3          for _version in versions:
   520   7810977    3811072.0      0.5      0.6              if _version == '__first_version__':
   521      3574       1834.0      0.5      0.0                  continue
   522   7807403  204602135.0     26.2     34.9              v = versions[_version]
   523   7807403  355139966.0     45.5     60.5              if v.attrs['prev_version'] == version_name:
   524      3573     749394.0    209.7      0.1                  v.attrs['prev_version'] = prev_version
   525      3574     360276.0    100.8      0.1          del versions[version_name]
   526
   527         1        178.0    178.0      0.0      versions.attrs['current_version'] = current_version
```

ArvidJB:
The previous implementation to compute `prev_version` for each version in `delete_versions` was accidentally quadratic: for each version to delete it would search over all current versions and fix their prev_version.
The new implementation finds the new `prev_version` for each version by walking backwards over `prev_version`.

Fixes #261.


ArvidJB:
What's the best way to add a speed test for this?

peytondmurray:
> What's the best way to add a speed test for this?

I'd be wary about actually trying to time it, instead I propose pulling out the contents of the loop to a separate function, `update_prev_version`. Then you can mock out `update_prev_version`, call the actual `update_prev_version` as a side effect, and use `MagicMock.call_count` to actually see if the number of times you needed to update an ancestor is what you'd expect. What do you think?

ArvidJB:
> > What's the best way to add a speed test for this?
> 
> I'd be wary about actually trying to time it, instead I propose pulling out the contents of the loop to a separate function, `update_prev_version`. Then you can mock out `update_prev_version`, call the actual `update_prev_version` as a side effect, and use `MagicMock.call_count` to actually see if the number of times you needed to update the `prev_version` is what you'd expect. What do you think?

I like the idea, but it's just one line we want to count:
```
            prev_version = versions[prev_version].attrs['prev_version']
```
I ended up using `sys.settrace` to count the number of times the line was executed.
What do you think, is that a too terrible hack?

asmeurer:
I wouldn't worry about adding any kind of speed test to the test suite but you can add something to the benchmarks.

peytondmurray (approved):
Implementation looks good to me. I'd like to revisit the test though, as checking a particular line will require maintenance for each change we make in `replay.py`. I'll make an issue to address this next.

Issue #198, opened 2021-07-06
Poor performance due to the default chunk size


gtca:
There seems to be an issue with the default chunk size used. At least, that seems to be an issue coming from `h5py`.

Here's a simple snippet:

```python
import jax
import h5py
from versioned_hdf5 import VersionedHDF5File

z = np.array(jax.random.normal(jax.random.PRNGKey(1), shape=(10_000_000,)))
with h5py.File("zdemo.h5", "w") as f:
    f["z"] = z
os.stat("xydemo.h5").st_size / 1e6
# => 1.202048

file = h5py.File("zdemo_versioned.h5", "w")
vf = VersionedHDF5File(file)
with vf.stage_version("v1") as f:
    f["z"] = z
file.close()
vf.close()

os.stat("zdemo_versioned.h5").st_size / 1e6
# => 40.538272
```

It then gets more complicated to build interfaces on top of versioned-hdf5 when there are 1D and multi-dimensional arrays and neither a single chunk size can be provided, nor the defaults can be used:

https://github.com/deshaw/versioned-hdf5/blob/09f3a8f6cccdaa9d5eb28811139e4a327e519079/versioned_hdf5/wrappers.py#L193-L196

This leads to a question if it might be possible to rely on `h5py` [to choose an appropriate chunk size](https://docs.h5py.org/en/stable/high/dataset.html#chunked-storage) rather than to define it with the `DEFAULT_CHUNK_SIZE`. That should also work beyond 1D arrays then.

asmeurer:
The default chunk size was chosen early on as something that appeared to have the best tradeoff between performance and file size, but we haven't investigated it in a while, so we might want to revisit it. 

The main issue is that due to the way Versioned HDF5 works, the chunk size has to remain constant across all versions of a dataset. That means that in general, you can't guess a good chunk size from the initial version of a dataset if its shape isn't representative of its typical shape across future versions. Also, the way it works is that dataset chunks are reused across versions, so the chunk size matters for more than just HDF5 read/write performance. It also directly affects how the dataset is stored in the versioned backend. In effect, all datasets work like they are set with `maxshape=(None, ..., None)`.

I think we could implement some sort of default chunk size for multidimensional datasets. It hasn't yet been something we've needed for our use-case, but it's something that would be convenient to do, although the caveats need to be documented.

gtca:
Thanks for your elaborate answer, @asmeurer!

[...]

PR #274, opened 2023-09-21
Guess a sensible default chunk size if none specified

peytondmurray:
This PR guesses a sensible default chunk size if none is provided, closing #198. Here, we make use of `h5py._hl.filters.guess_chunk` for this purpose. Note this only applies to 1D datasets, with the behavior of scalar and multidimensional datasets unchanged.

The change is quite short, but many of the tests were writing `DEFAULT_CHUNK_SIZE`-sized datasets to files, and then checking that the chunks written matched the `DEFAULT_CHUNK_SIZE`. As a result, many tests needed to be modified to lift this assumption.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -1,4 +1,5 @@
 import numpy as np
+from h5py._hl.filters import guess_chunk
 from h5py import VirtualLayout, VirtualSource, h5s
 from ndindex import Slice, ndindex, Tuple, ChunkSize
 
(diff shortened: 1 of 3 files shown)


asmeurer:
pytables has a lot of chunk size logic that they've investigated. https://www.pytables.org/usersguide/optimization.html

Something to be aware of is the tradeoff between performance and file size. Also the versioned-hdf5 usage of chunks might not always match the standard usage, so it's possible these other heuristics aren't quite correct.

peytondmurray:
After discussion, it sounds like we are good to merge.

Issue #273, opened 2023-09-21
delete_versions: use less memory and gracefully handle failure (PyInf#11013)


ArvidJB:
`delete_versions` recreates the raw dataset by creating a temporary `_tmp_raw_data` dataset. To populate `_tmp_raw_data` it creates a temporary numpy array in memory which is large enough to contain all data across all versions (in other words, it needs a lot of memory). This can fail:
```
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 515, in delete_versions
    _delete_dataset(f, name, versions_to_delete)
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 439, in _delete_dataset
    raw_data_chunks_map = _recreate_raw_data(f, name, versions_to_delete)
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 218, in _recreate_raw_data
    n = np.full(new_raw_data.shape, _get_np_fillvalue(raw_data), dtype=new_raw_data.dtype)
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/numpy/core/numeric.py", line 343, in full
    a = empty(shape, dtype, order)
MemoryError: Out of memory:
  Unable to allocate 1.16 GiB for an array with shape (124610, 1250) and data type float64
```

We tried to run `delete_versions` again, but now it failed because the `_tmp_raw_data` dataset is already present.
```
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 515, in delete_versions
    _delete_dataset(f, name, versions_to_delete)
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 439, in _delete_dataset
    raw_data_chunks_map = _recreate_raw_data(f, name, versions_to_delete)
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/versioned_hdf5/replay.py", line 208, in _recreate_raw_data
    new_raw_data = f['_version_data'][name].create_dataset(
  File "/usr/local/python/python3/std/lib64/python3.10/site-packages/h5py/_hl/group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
[...]

PR #276, opened 2023-09-27
Guard against exceptions when reconstructing raw data in `delete_versions`

peytondmurray:
This PR guards against exceptions raised when recreating the raw data during a call to `delete_versions`. This step allocates a potentially large numpy array, which may cause an OOM error. In this case, the `_tmp_raw_data` is deleted before the exception is raised.

An additional check is made to remove any existing `_tmp_raw_data` datasets before trying to reconstruct the raw data.

Partially addresses #273.

### DIFF
--- a/versioned_hdf5/replay.py
+++ b/versioned_hdf5/replay.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import List, Iterable, Union, Dict, Any
+from typing import List, Iterable, Union, Dict, Any, Optional
 from h5py import (
     VirtualLayout,
     h5s,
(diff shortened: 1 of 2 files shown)


versioned_hdf5/replay.py
```diff
@@ -215,35 +234,60 @@ def _recreate_raw_data(
...
-    if not tmp:
-        del f['_version_data'][name]['raw_data']
-        f['_version_data'][name].move('_tmp_raw_data', 'raw_data')
 
-    return raw_data_chunks_map
+    # Guard against existing _tmp_raw_data
+    _delete_tmp_raw_data(f, name)
+
+    try:
+        new_raw_data = f['_version_data'][name].create_dataset(
```
ArvidJB, line 242: I think the `create_dataset` call should be outside of the `try`, so:
```
new_raw_data = ...create_dataset(...)
try:
    ...
except Exception:
    _delete_tmp_raw_data(f, name)
```
I see that you are checking if `_tmp_raw_data` exists in `_delete_tmp_raw_data` (and we should keep that) but I think this moving `create_dataset` out is cleaner.

ArvidJB (approved):
LGTM, small suggestion!

versioned_hdf5/replay.py
peytondmurray: :100: agree, I've made that change. Thanks for the review!

Issue #216, opened 2022-02-01
h5py reference counting bug with release 1.3.7


ArvidJB:
With the changes in 1.3.7 we are sporadically getting the following error:
```
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/codemill/bessen/versioned_hdf5_env/lib64/python3.7/site-packages/h5py/_hl/files.py", line 461, in __exit__
    self.close()
  File "/codemill/bessen/versioned_hdf5_env/lib64/python3.7/site-packages/h5py/_hl/files.py", line 431, in close
    id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 269, in h5py.h5f.get_obj_ids
RuntimeError: Can't increment id ref count (can't locate ID)
```

I have tracked this down to this change
https://github.com/deshaw/versioned-hdf5/commit/e2509ae1624bb60d72b56bc028c3328ce78eeadb
which changes the way `Hashtable` is cached.

1. Why did we need the `lru_cache` in the first place? I assume this addresses some performance problem?
2. Is caching anything containing a reference to the `h5py.File` `f` invalid anyway?

Below is a reproducer which seems to cause this to happen with high probability:
```
import h5py
import numpy as np
from versioned_hdf5 import VersionedHDF5File

ds_names = ['ds{}'.format(i) for i in range(10)]
num_versions = 101


def f1():
    with h5py.File('foo.h5', 'w') as f:
        vf = VersionedHDF5File(f)
        with vf.stage_version('0') as sv:
            data_group = sv.create_group('data')
            for name in ds_names:
                data_group.create_dataset(name, data=np.arange(4))


def f2():
    for i in range(1, num_versions):
        with h5py.File('foo.h5', 'r+') as f:
            vf = VersionedHDF5File(f)
            with vf.stage_version(str(i)) as sv:
                data_group = sv['data']
                for name in ds_names:
                    ds = data_group[name]
                    ds.resize((i + 4,))
                    ds[:] = np.arange(i, 2 * i + 4)


def f3():
    with h5py.File('foo.h5', 'r') as f:
[...]

Issue #162, opened 2021-02-17
Iterating over versions leads to id error 


ArvidJB:
Not sure what's going on here - is this a reference counting problem perhaps? This does not seem to fail every time I run it, but fails 90% of the time?
```
In [1]: from versioned_hdf5 import VersionedHDF5File

In [2]: with h5py.File('foo.h5', 'w') as f:
   ...:     vf = VersionedHDF5File(f)
   ...:     with vf.stage_version('0') as sv:
   ...:         sv.create_dataset('bar', data=np.zeros(0, dtype='double'))
   ...:
[PYFLYBY] import h5py
[PYFLYBY] import numpy as np

In [3]: for i in range(1, 100):
   ...:     with h5py.File('foo.h5', 'r+') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         with vf.stage_version(str(i)) as sv:
   ...:             sv['bar'].resize((i,))
   ...:             sv['bar'][i-1] = i
   ...:

In [4]: with h5py.File('foo.h5', 'r') as f:
   ...:     versions = list(f['_version_data/versions'].keys())
   ...:

In [5]: with h5py.File('foo.h5', 'r') as f:
   ...:     vf = VersionedHDF5File(f)
   ...:     for version in versions:
   ...:         if version != '__first_version__':
   ...:             cv = vf[version]
   ...:             cv['bar'][:]
   ...:
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-5-73795d691ba0> in <module>
      4         if version != '__first_version__':
      5             cv = vf[version]
----> 6             cv['bar'][:]
      7

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

/usr/local/python/python-3.7/std/lib64/python3.7/site-packages/h5py/_hl/files.py in __exit__(self, *args)
    459     def __exit__(self, *args):
    460         if self.id:
--> 461             self.close()
    462
    463     @with_phil

/usr/local/python/python-3.7/std/lib64/python3.7/site-packages/h5py/_hl/files.py in close(self)
    429                 # Close file-resident objects first, then the files.
    430                 # Otherwise we get errors in MPI mode.
--> 431                 id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
[...]

PR #277, opened 2023-10-01
Delete versions OOM fix

peytondmurray:
This PR avoids writing a `_tmp_raw_data` array in calls to `_recreate_raw_data`, preventing an allocation that in the past could consume lots of memory, causing `delete_versions` to fail. Closes https://github.com/deshaw/versioned-hdf5/issues/273.

NOTE: This is a WIP PR - right now I'm seeing a really weird issue with the tests in `test_replay.py` where calls to `np.testing.assert_equal` are failing on tests which have had deleted versions. I'm guessing this is some issue flushing to disk; creating this PR here for further investigation.

Okay, confirmed that closing and opening the file objects in the offending tests seems to resolve the issue. Will look for a way to flush output to disk to close this out.

---

The intermittent test failures are indeed resolved by calling `gc.collect` at the end of `delete_versions`. Clearly this isn't ideal, but at the moment this seems like a functioning approach at least.

## Changes

- `_recreate_raw_data` now directly writes the recreated raw data to the file. Before, we were writing to a numpy array, then an intermediate `_tmp_raw_data` dataset, then we'd move the `_tmp_raw_data` to `raw_data` to replace the original raw data. This should resolve any issues related to memory usage, and will likely also speed up `delete_versions` overall, although we now pay an additional garbage collection penalty.
- Removed the `posixpath` alias, which was `pp` and thus would interfere with debugger `pp` pretty-print commands, in.

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -93,10 +93,17 @@ class VersionedHDF5File:
                     f"than what is currently installed. Please update versioned-hdf5."
                 )
 
-        self._version_data = f['_version_data']
-        self._versions = self._version_data['versions']
         self._closed = False
         self._version_cache = {}
+
+    @property
+    def _versions(self):
+        """Shorthand reference to the versions group of the file."""
+        return self.f['_version_data']['versions']
+
+    @property
+    def _version_data(self):
+        return self.f['_version_data']
 
     @property
     def closed(self):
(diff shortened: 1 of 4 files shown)


peytondmurray:
So a little more information about this, since it's still a mystery why some of the tests are intermittently failing on this branch:

I've got the right chunks being written to the right place, but I think I've uncovered a bug somewhere in the wrappers. On [this line](https://github.com/peytondmurray/versioned-hdf5/blob/141397b8ce8c27f219a4f464c26a18103232adc5/versioned_hdf5/wrappers.py#L92) of `InMemoryGroup.__getitem__`, if you index the entire array before returning the result, the tests pass. In other words, changing this

```python
        if name in self._data:
            return self._data[name]
```

to


```python
        if name in self._data:
            self._data[name][:]
            return self._data[name]
```

causes all the intermittently failing tests to pass. I'll keep investigating here.

ArvidJB:
> [...] This is incredibly weird, but points to some issue with caching. I'll keep investigating here.

Thanks for keeping us posted @peytondmurray. I agree, this looks like some issue with caching.

If you look at some of the other issues in the repo you can find that we ran into similar issues like this before without figuring out what was happening, like this one https://github.com/deshaw/versioned-hdf5/issues/216, or this one https://github.com/deshaw/versioned-hdf5/issues/162. We "fixed" the problem by calling `gc.collect`, does that help here?

peytondmurray:
This could definitely be a gc issue given how random it feels, so let me try doing this.

Issue #275, opened 2023-09-27
Bug in _rebuild_hashtables (PyInf#11013)


ArvidJB:
This commit https://github.com/deshaw/versioned-hdf5/commit/a14c496567aad3de279933cf8177be204a3c005b introduced a bug where old files (created with an old versioned_hdf5 version) containing "nested" datasets (i.e., in a group) would break when read with the latest versioned_hdf5 code.

Create a nested dataset with an older versioned_hdf5 version:
```
with h5py.File('/u/bessen/tmp/data.h5', 'w') as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version('r0') as sv:
        sv.create_group('data')
        sv['data/values'] = np.array([1, 2, 3])
```

Now try to read it with the latest versioned_hdf5 version:
```
In [9]: with h5py.File('/u/bessen/tmp/data.h5', 'r+') as f:
   ...:     vf = VersionedHDF5File(f)
   ...:
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[9], line 2
      1 with h5py.File('/u/bessen/tmp/data.h5', 'r+') as f:
----> 2     vf = VersionedHDF5File(f)

File /usr/local/python/python-3.10/std/lib64/python3.10/site-packages/versioned_hdf5/api.py:87, in VersionedHDF5File.__init__(self, f)
     82     if f.mode == 'r+':
     83         logger.info(
     84             f'Rebuilding hash tables for {f.filename}.',
     85             stacklevel=2
     86         )
---> 87         self._rebuild_hashtables()
     88         self.f['_version_data']['versions'].attrs['data_version'] = DATA_VERSION
     90 elif self.data_version_identifier > DATA_VERSION:

File /usr/local/python/python-3.10/std/lib64/python3.10/site-packages/versioned_hdf5/api.py:368, in VersionedHDF5File._rebuild_hashtables(self)
    366 for name in self.f['_version_data'].keys():
    367     if name != 'versions':
--> 368         del self.f['_version_data'][name]['hash_table']
    369         Hashtable.from_versions_traverse(self.f, name)

File h5py/_objects.pyx:54, in h5py._objects.with_phil.wrapper()

File h5py/_objects.pyx:55, in h5py._objects.with_phil.wrapper()

File /usr/local/python/python-3.10/std/lib64/python3.10/site-packages/h5py/_hl/group.py:489, in Group.__delitem__(self, name)
    486 @with_phil
[...]

PR #278, opened 2023-10-04
Fix issue when rebuilding hashtables of nested datasets

peytondmurray:
This PR fixes an issue where nested datasets (i.e. datasets which were named something like `_version_data/<name>/<othername>`) were unable to have their hash tables rebuilt correctly. Fixes https://github.com/deshaw/versioned-hdf5/issues/275.


ArvidJB:
Does this fix arbitrarily nested groups/datasets? If a Dataset is at `/foo/bar/baz/values` we should also handle that.

peytondmurray:
It should do, since datasets are recursively searched for. I'll make another test to be sure.

Issue #271, opened 2023-09-13
Rework `test_delete_versions_speed` to use mock


peytondmurray:
Currently, `test_delete_versions_speed` checks the number of executions of a particular line of code that deletes versions from a file. The current implementation requires that we update the line number every time a change is made to `replay.py`. I'd like to rework this so that we use `Mock.call_count` instead.

PR #281, opened 2023-10-16
Rewrite test_delete_versions_speed to use mocks

peytondmurray:
This PR rewrites the `test_delete_versions_speed` to use a mock instead of `linecache` and `sys.settrace`. This avoids us having to change the expected line number every time we make a change to `replay.py`.

I had to redo the math on the execution count here. Closes https://github.com/deshaw/versioned-hdf5/issues/271.


ArvidJB (approved):
That looks much better than the hacky code it replaces!

Issue #280, opened 2023-10-15
Corrupted hash_table after rebuilding - PyInf#11257


ArvidJB:
When a file is upgraded to the newer `data_version` the `hash_table` gets corrupted.

I created the file with version 1.3.14:
```
In [1]: with h5py.File('data.h5', 'w') as f:
   ...:     vf = VersionedHDF5File(f)
   ...:     with vf.stage_version(str(0)) as sv:
   ...:         sv['foo'] = np.arange(0)
   ...:
[PYFLYBY] from versioned_hdf5 import VersionedHDF5File
[PYFLYBY] import h5py
[PYFLYBY] import numpy as np

In [2]: for i in range(1, 11):
   ...:     with h5py.File('data.h5', 'r+') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         with vf.stage_version(str(i)) as sv:
   ...:             sv['foo'] = np.arange(i)
   ...:

In [3]: f = h5py.File('data.h5', 'r')

In [4]: f['_version_data']['foo/hash_table'][:]
Out[4]:
array([([244, 110, 238,  27, 224,  41,  59, 223,  96,   3,  54, 188, 189, 182,  70,  16,   0,  64, 130, 211,  17,  73, 196,  96, 137, 183, 173, 134, 219,  60, 163, 225], [    0,     1]),
       ([ 60,  83,  79,  25,  69, 142, 253, 190,  52,   0, 238, 145, 202, 172, 248,  37, 221, 170,  99, 152, 128, 140, 253, 255, 133, 207,   1, 186, 120, 206,  21,   4], [ 4096,  4098]),
       ([107,  30,   1, 112, 223,  56,  54,  73,  31, 252, 250,   2,  94, 196,  67, 127, 163, 234, 136,  51, 146, 207, 127, 197, 253, 135, 151, 164, 218,   1, 177, 242], [ 8192,  8195]),
       ([227, 104, 150, 169, 229, 121, 142, 237,  80,  44, 100,  20, 111,  54, 121, 103,  92, 248,  71, 238, 233, 191, 189, 186, 142,  68, 237, 230, 198, 131, 117,  79], [12288, 12292]),
       ([ 58, 184,  70,  41, 133, 109, 166, 145, 140, 224, 123, 221, 227,  90, 194, 204, 175, 232, 173,  92, 200, 150,  63, 190,  71, 119,  16, 250, 213,  16, 200, 186], [16384, 16389]),
[...]

PR #282, opened 2023-10-18
Fix hashtable rebuilding logic to not corrupt hashtables

ArvidJB:
The changes in https://github.com/deshaw/versioned-hdf5/commit/b0e6d19a197a6748191bda8d5f229545cb4b62a0
unfortunately corrupted the hashtables by picking raw chunks
in increasing order rather than picking the correct raw chunk.

Note that this broke *all* hashtables, not just dtype='O'
hashtables which were the target of that fix. Because of
that all `data_version == 2` files are corrupted.

This commit fixes the rebuild hashtable logic by actually
looking up the raw data slice for each chunk. Furthermore
1. rebuilding is now a separate command that needs to be
   called implicitly, rather than it happening silently.
2. if it encounters a `data_version == 2` file it raises
   an exception and asks the user to recreate the file.
3. it only rebuilds hashtables for dtype='O'.

Fixes https://github.com/deshaw/versioned-hdf5/issues/280.


ArvidJB:
This needs more tests:
* add write chunk reuse tests to other rebuild tests
* multi-dimensional arrays
* non-ascii unicode characters

peytondmurray (approved):
@ArvidJB Thanks for looking into this! I'm glad there's a path forward where the hash tables can be fixed - I know in the issue you mentioned it might just be worth making old data versions incompatible.

PR #287, opened 2023-10-30
Update error/warning message to give correct instructions

ArvidJB:
Fix error message since you cannot directly pass a file name to `VersionedHDF5File`.

Internal ticket: PyInf#11354


ArvidJB:
Merged as part of #288 .

Issue #288, opened 2023-11-07
Collisions when hashing unbounded strings


ArvidJB:
Because `dtype='O'` arrays of strings are hashed by computing a SHA256 hash of their contents we can get collisions if the concatenation of the contents of the array are the same:
```
>>> import h5py
... import versioned_hdf5
... import numpy as np
...
... path = "temp.h5"
...
... with h5py.File(path, mode="w") as f:
...     file = versioned_hdf5.VersionedHDF5File(f)
...     with file.stage_version("r0") as group:
...         group.create_dataset(
...             "values", shape=(2,), dtype=h5py.string_dtype(encoding='ascii'), data=np.array([b"a", b"bc"], dtype=object),
...             maxshape=(None,), chunks=(10000,), compression="gzip", shuffle=False,
...         )
...     with file.stage_version("r1") as group:
...         group.create_dataset(
...             "values", shape=(2,), dtype=h5py.string_dtype(encoding='ascii'), data=np.array([b"ab", b"c"], dtype=object),
...             maxshape=(None,), chunks=(10000,), compression="gzip", shuffle=False,
...         )
...
... with h5py.File(path, mode="r") as f:
...     file = versioned_hdf5.VersionedHDF5File(f)
...     for i in range(2):
...         print(file[f"r{i}"]["values"][:])
...
[b'a' b'bc']
[b'a' b'bc']
```
Here the two versions were supposed to contain `[b'a' b'bc']` and `[b'ab' b'c']`, but, because the concatenation of the strings is the same, the hashes are the same, and the chunk gets (incorrectly) reused.

ArvidJB:
I think we need to include the length of the string into the hash as well. Something like
```
     def hash(self, data: np.ndarray):
         """
         Compute hash for `data` array.
         """
         # Object dtype arrays store the ids of the elements, which may or may not be
         # reused, making it unsuitable for hashing. Instead, we need to make a combined
         # hash with the value of each element.
         if data.dtype == 'object':
             hash_value = self.hash_function()
             for value in data.flat:
                 if isinstance(value, str):
                     # default to utf-8 encoding since it's a superset of ascii (the only other
                     # valid encoding supported in h5py)
                     value = value.encode('utf-8')
+                hash_value.update(str(len(value)).encode('ascii'))
                 hash_value.update(value)
             hash_value.update(bytes(str(data.shape), 'utf-8'))
             return hash_value.digest()
         else:
             return self.hash_function(
[...]

PR #289, opened 2023-11-07
Include string length when hashing

ArvidJB:
When hashing `dtype=object` arrays we need to include the length of the individual strings, otherwise arrays with identical contents after concatenating all elements hash to the same value:
```
>>> import h5py
... import versioned_hdf5
... import numpy as np
...
... path = "temp.h5"
...
... with h5py.File(path, mode="w") as f:
...     file = versioned_hdf5.VersionedHDF5File(f)
...     with file.stage_version("r0") as group:
...         group.create_dataset(
...             "values", shape=(2,), dtype=h5py.string_dtype(encoding='ascii'), data=np.array([b"a", b"bc"], dtype=object),
...             maxshape=(None,), chunks=(10000,), compression="gzip", shuffle=False,
...         )
...     with file.stage_version("r1") as group:
...         group.create_dataset(
...             "values", shape=(2,), dtype=h5py.string_dtype(encoding='ascii'), data=np.array([b"ab", b"c"], dtype=object),
...             maxshape=(None,), chunks=(10000,), compression="gzip", shuffle=False,
...         )
...
... with h5py.File(path, mode="r") as f:
...     file = versioned_hdf5.VersionedHDF5File(f)
...     for i in range(2):
...         print(file[f"r{i}"]["values"][:])
...
[b'a' b'bc']
[b'a' b'bc']
```

By including the length of each string as part of the hash the hashes will now be unique.

This also includes the changes in https://github.com/deshaw/versioned-hdf5/pull/287 and supersedes that PR.

Fixes https://github.com/deshaw/versioned-hdf5/issues/288


versioned_hdf5/tests/test_hashtable.py
```diff
@@ -86,3 +86,29 @@ def test_object_dtype_hashes_values(tmp_path):
...
+                maxshape=(None,), chunks=(100,)
+            )
+        with file.stage_version("r1") as group:
+            group["values"] = np.array([b"ab", b"", b"cd"], dtype=object)
+        with file.stage_version("r2") as group:
+            group["values"] = np.array([b"ab", b"c", b"d"], dtype=object)
+
+    with h5py.File(filename, mode="r") as f:
+        file = VersionedHDF5File(f)
+        np.testing.assert_equal(file["r0"]["values"][:], np.array([b"a", b"b", b"cd"], dtype=object))
```
peytondmurray, line 112: Perfect, thank you!

peytondmurray (approved):
This looks great - thank you for this!

PR #296, opened 2023-11-30
Add check of hashed data when writing new data

### LINKED ISSUE #294: Hash collision checks
In light of our recent issues with hashing (see https://github.com/deshaw/versioned-hdf5/issues/280 and https://github.com/deshaw/versioned-hdf5/issues/288), we should add a simple sanity check when we write chunks. In particular I am thinking we should modify the logic in `write_dataset` here
https://github.com/deshaw/versioned-hdf5/blob/ee3efe0ded7bf175993bfef5a04308618a46c195/versioned_hdf5/backend.py#L128-L135
to add a check like
```
if raw_slice2 == raw_slice:
    slices_to_write[raw_slice] = s
elif not np.array_equal(ds[Tuple(raw_slice2, *[slice(0, i) for i in data_s.shape[1:]]).raw], data_s, equal_nan=True):
    raise ValueError('hashed data chunk and new data chunk do not match')
```

I'd argue that this check will be very fast (compared to the SHA256 computation at least!) so the extra overhead will be negligible.

peytondmurray:
This PR adds a check when writing data to a file. This check verifies that chunks of data that already exist in the hashtable correspond to arrays that match the data we are trying to write.

This should incur a performance overhead when writing datasets that reuse chunks, but ensures we won't be corrupting hashtables in the future. Closes #294.

Note: Currently I'm seeing a test failure locally with `test_delete_string_dataset` on `master`. The failure happens when trying to open the file after running `h5repack` on the `hdf5` file, with `OSError: Unable to open file (bad object header version number)`. I'm opening this PR to see if CI sees it.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -1,4 +1,7 @@
+from typing import Dict, Optional
+
 import numpy as np
+from numpy.testing import assert_array_equal
 from h5py._hl.filters import guess_chunk
 from h5py import VirtualLayout, VirtualSource, h5s
 from ndindex import Slice, ndindex, Tuple, ChunkSize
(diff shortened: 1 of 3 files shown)


ArvidJB:
The same check should also go into `write_dataset_chunks`.

Can you add a test where the hash function is mocked to be one of the previous data_version hash functions? The check should get triggered.

peytondmurray:
Looks like there must be a problem locally with `h5repack` that is responsible for the failing test I'm seeing, and anyway it isn't a failing test related to this change.

---

## Changes

- Removed a little dead code
- Changed the logic slightly to make it more readable by removing the usage of `setdefault`. This let me get the clarity I needed to figure out how to carry out the check of the hashed data.
- Added check of the hashed data
- Moved some code inside the branch statement that only needed to be run on one branch

peytondmurray:
Okay, I think we have our bases covered with these tests.

versioned_hdf5/backend.py
```diff
@@ -124,23 +127,143 @@ def write_dataset(f, name, data, chunks=None, dtype=None, compression=None,
...
+
+    Returns
+    -------
+    np.ndarray
+        Object dtype array filled with elements of type bytes
+    """
+    if len(arr) > 0 and isinstance(arr[0], bytes):
+        return arr
+    else:
+        return np.vectorize(lambda i: bytes(i, encoding='utf-8'))(arr)
```
ArvidJB, line 264: You can use `np.char.encode` here, I think?

ArvidJB (approved):
Looks good to me!

versioned_hdf5/backend.py
peytondmurray: Normally I definitely would but this doesn't work for object dtype arrays:

image

Issue #298, opened 2023-12-12
delete_versions high memory usage (PyInf#11623)


ArvidJB:
We observed that a number of our processes were running out of memory while running `delete_versions`. It turned out that we ended up choosing specific chunk sizes which inadvertently created a lot more chunks than ideal. Nonetheless it should be possible to reduce the memory consumed!

Here's some code to reproduce this. It creates a file with 100 versions, each version ended up having 35889 chunks:
```
import contextlib
import logging
import os
import resource
import shutil
import tempfile

import h5py
import numpy as np
import psutil

from   versioned_hdf5           import VersionedHDF5File, delete_versions


@contextlib.contextmanager
def memory_limit_ctx(buffer):
    # We want to set the soft limit to be the current usage + buffer.
    assert buffer > 0
    # get current memory usage
    used = psutil.Process(os.getpid()).memory_info().rss
    old_limit = resource.getrlimit(resource.RLIMIT_AS)
    new_limit = (used + buffer, old_limit[1])
    try:
        resource.setrlimit(resource.RLIMIT_AS, new_limit)
        yield
    finally:
        resource.setrlimit(resource.RLIMIT_AS, old_limit)


if __name__ == '__main__':
    # use an existing file with 100 versions
    filename = 'delete_versions_memory_usage.h5'

    # # to create the file uncomment on first run
    # logging.getLogger().setLevel(logging.INFO)
    # with h5py.File(filename, 'w') as f:
    #     vf = VersionedHDF5File(f)
    #     logging.info(f'creating version r{0}')
    #     with vf.stage_version(f'r0') as sv:
    #         sv.create_dataset('bitmap', data=np.zeros((35889, 5), dtype=np.uint8), compression='lzf',
    #                           dtype=np.dtype(np.uint8), maxshape=(None, None), chunks=(1, 10000))
    # with h5py.File(filename, 'r+') as f:
    #     vf = VersionedHDF5File(f)
    #     for i in range(1, 100):
    #         logging.info(f'creating version r{i}')
    #         with vf.stage_version(f'r{i}') as sv:
    #             sv['bitmap'][:] = np.random.randint(0, 256, 35889 * 5, dtype='uint8').reshape((35889, 5))

    # now delete some versions in place
    tmp_name = tempfile.mktemp('.h5')
    shutil.copy2(filename, tmp_name)
    try:
        # want to keep only every other version
        versions_to_delete = []
        with h5py.File(tmp_name, 'r') as f:
[...]

PR #300, opened 2024-01-08
Optimize `_recreate_raw_dataset` to use less memory; add versions method

peytondmurray:
This PR optimizes a part of the `_recreate_raw_dataset` that currently can allocate significant memory for datasets with large numbers of versions. I also saw a large speedup in this section of the code post-refactor.

A `VersionedHDF5File.versions` method was added (a `VersionedHDF5` should know what versions it has), which should return the same thing as `versioned_hdf5.versions.all_versions` but be much less verbose and more convenient.

Partially addresses https://github.com/deshaw/versioned-hdf5/issues/298. Deleting datasets can still be costly in terms of memory because the operation requires 

1. Reconstructing the raw dataset, which entails keeping track of the raw chunks that need to be kept
2. Reconstructing the hashtable, which requires loading the inverse of the old hashtable as a new one gets built
3. Recreating the virtual datasets for the versions of the dataset we are keeping

### DIFF
--- a/versioned_hdf5/api.py
+++ b/versioned_hdf5/api.py
@@ -575,3 +575,17 @@ class VersionedHDF5File:
                 diff[vir1] = (self[v1][name][vir1.raw], None)
 
         return diff
+
+    @property
+    def versions(self) -> List[str]:
+        """Return the names of the version groups in the file.
+
+        This should return the same as calling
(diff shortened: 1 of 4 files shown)


peytondmurray:
This change is low risk; merging now.

PR #313, opened 2024-02-22
Detect cases where unused chunk space can be written to

### LINKED ISSUE #295: Detect appends and reuse chunks (PyInf#10777)
A lot of our writes are actually just appends to existing data. Currently all appends to a chunk are treated the same as any other update: resizing or changing a single value results in a completely new chunk being written to the raw data.

But for pure appends which we do not change any of the existing values we can do better. Take this example below which ends up creating a new raw chunk for each version written:
```
>>> # THIS CELL WAS AUTO-GENERATED BY PYFLYBY
>>> import h5py
>>> import numpy as np
>>> import shutil
>>> from tempfile       import mkdtemp
>>> from versioned_hdf5 import VersionedHDF5File
>>> # END AUTO-GENERATED BLOCK


>>> d = mkdtemp()


>>> chunks = (10,)
>>> num_updates = 15


>>> with h5py.File(d + '/data.h5', 'w') as f:
...     vf = VersionedHDF5File(f)
...     with vf.stage_version('r01') as sv:
...         sv.create_dataset('values', data=np.array([0]), chunks=chunks)
...     for i in range(2, num_updates + 1):
...         with vf.stage_version(f'r{i:02d}') as sv:
...             sv['values'].resize((i,))
...             sv['values'][:] = np.arange(i)


>>> with h5py.File(d + '/data.h5', 'r') as f:
...     value_raw_data = f['_version_data/values/raw_data']
...     print(value_raw_data[:].reshape((-1,) + value_raw_data.chunks))

[[ 0  0  0  0  0  0  0  0  0  0]
 [ 0  1  0  0  0  0  0  0  0  0]
 [ 0  1  2  0  0  0  0  0  0  0]
 [ 0  1  2  3  0  0  0  0  0  0]
 [ 0  1  2  3  4  0  0  0  0  0]
 [ 0  1  2  3  4  5  0  0  0  0]
 [ 0  1  2  3  4  5  6  0  0  0]
 [ 0  1  2  3  4  5  6  7  0  0]
 [ 0  1  2  3  4  5  6  7  8  0]
 [ 0  1  2  3  4  5  6  7  8  9]
 [10  0  0  0  0  0  0  0  0  0]
 [10 11  0  0  0  0  0  0  0  0]
 [10 11 12  0  0  0  0  0  0  0]
 [10 11 12 13  0  0  0  0  0  0]
 [10 11 12 13 14  0  0  0  0  0]]


>>> from versioned_hdf5.slicetools import spaceid_to_slice
>>> with h5py.File(d + '/data.h5', 'r') as f:
...     for version in f['_version_data/versions']:
...         if version == '__first_version__':
...             continue
...         values_vds = f['_version_data/versions'][version]['values']
...         chunk_map = ', '.join([f'{spaceid_to_slice(vs.vspace)} -> {spaceid_to_slice(vs.src_space)}' for vs in values_vds.virtual_sources()])
...         print(f'{version}: {chunk_map}')

r01: Tuple(slice(0, 1, 1)) -> Tuple(slice(0, 1, 1))
r02: Tuple(slice(0, 2, 1)) -> Tuple(slice(10, 12, 1))
r03: Tuple(slice(0, 3, 1)) -> Tuple(slice(20, 23, 1))
r04: Tuple(slice(0, 4, 1)) -> Tuple(slice(30, 34, 1))
r05: Tuple(slice(0, 5, 1)) -> Tuple(slice(40, 45, 1))
r06: Tuple(slice(0, 6, 1)) -> Tuple(slice(50, 56, 1))
r07: Tuple(slice(0, 7, 1)) -> Tuple(slice(60, 67, 1))
r08: Tuple(slice(0, 8, 1)) -> Tuple(slice(70, 78, 1))
r09: Tuple(slice(0, 9, 1)) -> Tuple(slice(80, 89, 1))
r10: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1))
r11: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1)), Tuple(slice(10, 11, 1)) -> Tuple(slice(100, 101, 1))
r12: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1)), Tuple(slice(10, 12, 1)) -> Tuple(slice(110, 112, 1))
r13: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1)), Tuple(slice(10, 13, 1)) -> Tuple(slice(120, 123, 1))
r14: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1)), Tuple(slice(10, 14, 1)) -> Tuple(slice(130, 134, 1))
r15: Tuple(slice(0, 10, 1)) -> Tuple(slice(90, 100, 1)), Tuple(slice(10, 15, 1)) -> Tuple(slice(140, 145, 1))


>>> shutil.rmtree(d)
```

But here we can be more clever since those versions do not fill the chunk completely. We can change the raw chunk by writing past the previous slice's end!

In this case we could just end up using only two raw chunks: `[ 0  1  2  3  4  5  6  7  8  9]` and `[10 11 12 13 14  0  0  0  0  0]`!

It will most likely be impossible to detect general chunk reuse possibilities. But for the special case where a Dataset is resized we can detect if the *last chunk* can be reused by comparing the chunk's slice matching the existing slice.

peytondmurray:
This PR adds an `.append()` method to allow `InMemoryDataset` objects to append data to any unused space in the last chunk. If no free space exists, data is written as usual into a new chunk.

Closes #295.

## Changes

- Refactored the `data_dict` system that previously kept track of the slices in an `InMemoryDataset`. Previously the `data_dict` was a member which lived on the low level `InMemoryDatasetID` object, and contained mappings between slices in the virtual dataset and slices in the raw dataset. After careful consideration, it doesn't seem necessary to keep the data on this lower level object, we can just keep it on the InMemoryDataset, significantly reducing the conceptual complexity of the data model.

Previously, when the user called `InMemoryDataset.resize` or `InMemoryDataset.__setitem__`, the `data_dict` mapping would keep track of slices that changed as they changed; when exiting the `stage_version` context manager for a version, this `data_dict` was then used to both

1. Write new data to the underlying `raw_data` referenced by the virtual dataset
2. Provide the virtual slices that made up the new version of the dataset

This system has been replaced with a more explicit scheme with defers all computation until execution exits the `stage_version` context manager. With the new scheme, the user can manipulate data in three ways:

1. `InMemoryDataset.__setitem__`
2. `InMemoryDataset.resize`
3. `InMemoryDataset.append`

In all cases, when the user calls one of these methods the `InMemoryDataset` keeps track of the manipulation done by the user. When the `stage_version` context is exited, all computations are resolved then. Therefore we should expect to see some of the computation time that would otherwise be spent inside the context manager to be reduced, but with a corresponding increase when the context manager is exited.

- `InMemoryDataset` now keeps track of the last element written to its corresponding `raw_data`, which is needed for `append` operations
- `backend.create_virtual_dataset` can now accept either `Slice` objects or `Tuple` objects in the values of the `slices` parameter
- Added `SetOperation`, `ResizeOperation`, and `AppendOperation` helper classes which are used to keep track of the deferred operations on `InMemoryDataset` objects
- Added an `AppendChunk` class to help with deferring writes when the user calls `.append`. A similar scheme would simplify the code for `__setitem__` and `resize` calls, but I've left that refactor for another PR.
- Added a `partition` method which partitions a `ndindex.Tuple` or `np.ndarray` object into chunks of a requested size. I realize that there's already an `ndindex` function which does this, but the syntax is arcane. Parts of this PR took a long time to comprehend, so I'm trying to reduce the maintenance burden going forward.
- Added a few functions in `slicetools.py`:
  - `to_slice_tuple`: convert `Tuple` of arbitrary `ndindex` types to `Tuple` of `Slice` types, so that `obj.args[0].start` etc doesn't fail
  - `to_raw_index`: convert a relative (virtual) index to an index in a raw chunk
  - `get_vchunk_overlap`: helper to get the overlap of an arbitrary virtual chunk and an index into a virtual dataset
  - `get_shape`: helper which gets the size of an index along each dimension

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -26,6 +26,7 @@ py.install_sources(
     'versioned_hdf5/hashtable.py',
     'versioned_hdf5/replay.py',
     'versioned_hdf5/slicetools.py',
+    'versioned_hdf5/utils.py',
     'versioned_hdf5/versions.py',
     'versioned_hdf5/wrappers.py',
   ],
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,3 +4,4 @@ doctest_optionflags= NORMALIZE_WHITESPACE
 markers =
     setup_args : kwargs for setup fixture.
     slow: slow tests
+    append: tests of the append machinery
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -1,18 +1,205 @@
-from typing import Dict, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 import numpy as np
-from h5py import VirtualLayout, VirtualSource, h5s
+from h5py import Dataset, File, Group, VirtualLayout, VirtualSource, h5s
 from h5py._hl.filters import guess_chunk
 from ndindex import ChunkSize, Slice, Tuple, ndindex
 from numpy.testing import assert_array_equal
 
 from .hashtable import Hashtable
+from .slicetools import (
+    get_shape,
+    get_vchunk_overlap,
+    partition,
+    spaceid_to_slice,
+    to_raw_index,
+    to_slice_tuple,
+)
+from .utils import AppendChunk, WriteChunk
 
 DEFAULT_CHUNK_SIZE = 2**12
 DATA_VERSION = 4
 # data_version 2 has broken hashtables, always need to rebuild
 # data_version 3 hash collisions for string arrays which, when concatenated, give the same string
 CORRUPT_DATA_VERSIONS = frozenset([2, 3])
+
+if TYPE_CHECKING:
+    from .wrappers import InMemoryDataset
+
+
+class SplitResult:
+    """Object which stores the result of splitting a dataset across the last chunk."""
+
+    def __init__(
+        self,
+        arr_to_append: np.ndarray,
+        arr_to_write: np.ndarray,
+        new_raw_last_chunk: Tuple,
+        new_raw_last_chunk_data: np.ndarray,
+    ):
+        """Init SplitResult.
+
+        Parameters
+        ----------
+        arr_to_append : np.ndarray
+            Array to be appended to the dataset
+        arr_to_write : np.ndarray
+            Array to be written into new chunks
+        new_raw_last_chunk : Tuple
+            Slice of the raw data containing the last chunk
+        new_raw_last_chunk_data : np.ndarray
+
+        """
+        self.arr_to_append = arr_to_append
+        self.arr_to_write = arr_to_write
+        self.new_raw_last_chunk = new_raw_last_chunk
+        self.new_raw_last_chunk_data = new_raw_last_chunk_data
+
+    def has_append_data(self) -> bool:
+        """Check whether there is data to append.
+
+        Returns
+        -------
+        bool
+            True if there is data to append, False otherwise
+        """
+        return self.arr_to_append.size > 0
+
+    def has_write_data(self) -> bool:
+        """Check whether there is data to write into a new chunk.
+
+        Returns
+        -------
+        bool
+            True if there is data to write into a new chunk, False otherwise
+        """
+        return self.arr_to_write.size > 0
+
+    def get_additional_rchunks_needed(self, chunk_size: int) -> int:
+        """Compute the number of additional chunks needed in the raw dataset.
+
+        Parameters
+        ----------
+        chunk_size : int
+            Number of elements in each chunk (along axis 0)
+
+        Returns
+        -------
+        int
+            Number of additional chunks of size chunk_size needed to store the data
+            that doesn't fit in the remaining free space in the dataset
+        """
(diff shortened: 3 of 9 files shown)


peytondmurray:
Thanks for the feedback, having an initial impression is useful here.

> Is it possible to break this huge change into multiple smaller changes? We can start by only appends.

One of the major issues I encountered here was figuring out where writes need to be made in memory vs to a file. If we decide to take a lazy approach above I think it won't work to just add the append method. But that brings us to your next point:

> I am not a huge fan of breaking up things into WriteOperation and delaying their application until committing. I think that introduces a lot of extra complexity to be able to handle all combinations.

Understandable, and you're probably right about this. Let me get back to you about how we can move forward before we start considering the other implications of this approach.

PR #315, opened 2024-03-13
Make create_virtual_dataset handle zero-length raw_data correctly

### LINKED ISSUE #314: Bug when creating empty dataset - PyInf#12125
We ran into a weird bug related to some ancient code in `create_virtual_dataset` where we create a virtual dataset of shape `(1,)` if the "raw_data" is empty:
https://github.com/deshaw/versioned-hdf5/blob/d6c709c36c039c6252ec7ed51fe411b32f8aa145/versioned_hdf5/backend.py#L400-L404
Normally this never happens since we would normally not get here if the dataset is empty. To reproduce this we need to create a non-empty dataset which later turns empty and then use `delete_versions` to delete all non-empty versions. A subsequent upsert then creates a corrupted empty virtual dataset:
```
import h5py
import numpy as np
from   versioned_hdf5           import VersionedHDF5File
from   versioned_hdf5.replay    import delete_versions

if __name__ == '__main__':
    path = '/var/tmp/12125.h5'
    with h5py.File(path, 'w') as f:
        vf = VersionedHDF5File(f)
            sv.create_dataset('values', data=np.array([1, 2, 3]))

    with h5py.File(path, 'r+') as f:
        vf = VersionedHDF5File(f)
            sv['values'].resize((0,))

    with h5py.File(path, 'r+') as f:
        delete_versions(f, ['r0'])

    with h5py.File(path, 'r+') as f:
        vf = VersionedHDF5File(f)
        with vf.stage_version('r2') as sv:
            sv['values'].resize((0,))

    with h5py.File(path, 'r+') as f:
        vf = VersionedHDF5File(f)
        cv = vf[vf.current_version]
        print(cv['values'][:])
```
gives the following error
```
  File "/codemill/bessen/opensource/versioned-hdf5/12125_empty_dataset_reproducer.py", line 29, in <module>
    print(cv['values'][:])
          ~~~~~~~~~~~~^^^
  File "/usr/local/python/python3/std/lib64/python3.11/site-packages/versioned_hdf5/wrappers.py", line 1129, in __getitem__
    return self.dataset.__getitem__(index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/usr/local/python/python3/std/lib64/python3.11/site-packages/versioned_hdf5/wrappers.py", line 658, in __getitem__
    return super().__getitem__(idx.raw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/usr/local/python/python3/std/lib64/python3.11/site-packages/h5py/_hl/dataset.py", line 768, in __getitem__
    return self._fast_reader.read(args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py/_selector.pyx", line 376, in h5py._selector.Reader.read
OSError: Can't synchronously read data (src and dest dataspaces have different number of elements selected)
```

peytondmurray:
## Summary

This PR makes `create_virtual_dataset` correctly handle the case where the raw data is empty, returning an empty virtual dataset.

Closes #314.

## Changes

- If the raw data for a dataset is empty, the virtual dataset is also empty instead of being shape `(1,)` (??)
- Added tests of both the backend (to see if zero-length raw datasets are correctly handled) and of the API (to cover the reproducer submitted with the original issue)
- Added ANN001 (missing type annotations) to the list of rules to ignore in for test functions. These are either parametrizations or pytest fixtures.

I'm still not entirely sure why it is necessary to close/open the file descriptor between each operation to reproduce this issue, but it must be something to do with flushing writes to the file. @ArvidJB any ideas?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,7 @@ ignore = [
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["F401"]
+"test_*.py" = ["ANN001"]
 
 [project.optional-dependencies]
 dev = ["pre-commit>=3.6.0"]
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -398,10 +398,7 @@ def create_virtual_dataset(
     slices = {c: s.reduce() for c, s in slices.items()}
 
     if len(raw_data) == 0:
-        shape = ()
-        layout = VirtualLayout((1,), dtype=raw_data.dtype)
-        vs = VirtualSource(".", name=raw_data.name, shape=(1,), dtype=raw_data.dtype)
-        layout[0] = vs[()]
+        layout = VirtualLayout(shape=(0,), dtype=raw_data.dtype)
     else:
         # Chunks in the raw dataset are expanded along the first dimension only.
         # Since the chunks are pointed to by virtual datasets, it doesn't make
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -2646,3 +2646,37 @@ def test_versions_property(vfile):
     # Delete some versions and check for the correct versions again
     delete_versions(vfile, versions_to_delete)
     assert set(all_versions(vfile.f)) == set(vfile.versions)
+
+
+def test_make_empty_dataset(tmp_path):
+    """Check that creating a dataset before making it empty can be done successfully.
+
+    This test would pass unless the file gets closed/reopened for each operation,
+    which is why we do that here; unsure about why that is, but it must be related to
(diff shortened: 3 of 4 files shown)


peytondmurray:
Merging this as is to close out this bug. If any suspicious behavior persists, we can open a new issue.

PR #324, opened 2024-05-22
Faster and correct chunk reuse logging

### LINKED ISSUE #323: Overhead in _log_version_diff_stats (PyInf#12680)
It looks like the logging in `_log_version_diff_stats` is not correct and also pretty slow.
* Not correct, because it compare the hashtable against the hashtable, incorrectly concluding that on every commit all chunks were reused.
* Slow, because it ends up loading the entire hashtable, irrespective of
   * whether the dataset was actually modified
   * which chunks were actually written

ArvidJB:
Closes #323.

The logic in `_log_version_diff_stats` is currently not correct since it just compares the new `hashtable` with itself. This ends up always logging that all chunks were reused, which is obviously not correct. Furthermore, the size of the hashtable also does not necessarily correlate with the number of chunks reused for a dataset version.

It is also fairly slow since loading the hashtable is expensive if it drops out of the lru_cache (and unnecessary anyway, as discussed above). The calculations also happen regardless of the logging level, which means that they are mostly wasted anyway.

This change simplifies the logging of version diff stats by doing it directly when the dataset is written and when we can observe the number of added reused/chunks.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Dict, Optional
 
 import numpy as np
@@ -175,6 +176,9 @@ def write_dataset(
     chunk_size = chunks[0]
 
     with Hashtable(f, name) as hashtable:
+        old_chunks = hashtable.largest_index
+        chunks_reused = 0
+
         if len(data.shape) != 0:
             for data_slice in ChunkSize(chunks).indices(data.shape):
                 data_s = data[data_slice.raw]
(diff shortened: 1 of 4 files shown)


versioned_hdf5/backend.py
```diff
@@ -331,12 +347,10 @@ def write_dataset_chunks(f, name, data_dict, shape=None):
     chunks = tuple(raw_data.attrs["chunks"])
     chunk_size = chunks[0]
 
-    if shape is None:
```
peytondmurray, line 334: Thanks for this!

peytondmurray (approved):
Thanks for this - looks good to me.

PR #326, opened 2024-05-25
Optimize `InMemoryDataset.resize` to avoid quadratic performance

peytondmurray:
This PR optimizes `InMemoryDataset.resize` to avoid quadratic performance with the number of chunks in `InMemoryDataset.data_dict`; closes #325.

When calling `resize()`, any chunk of the dataset which is at the edge of the new dataset's shape needs to be read, and these chunks then need to be written to the raw dataset. In the process of reading the data for each edge-chunk in `Dataset.data_dict`, each call to `InMemoryDataset.__getitem__` would execute `InMemoryDatasetID.can_read_direct`, which itself iterates through every item in `Dataset.data_dict`, leading to very slow performance for certain datasets with large `data_dict`.

This PR introduces a new `Dataset.get_index` method which takes a `can_read_direct` parameter, allowing `InMemoryDataset.resize` to call `InMemoryDatasetID.can_read_direct` just once per resize, effectively eliminating this overhead. The existing `InMemoryDataset.__getitem__` method itself calls `InMemoryDataset.get_index`.

This PR also adds a test to ensure that `resize` only results in one call to `can_read_direct` to ensure this doesn't happen again. Docstrings and type annotations were also improved.

Issue #325, opened 2024-05-24
resize is very expensive ("quadratic" performance) (PyInf#12702)


ArvidJB:
The `.resize()` call is expensive in `versioned-hdf5` because it has "quadratic" performance. A `resize` call
1. creates a complete `data_dict` for the existing dataset and then iterates over each of the chunks
2. for each new chunk in the iteration which partially overlaps with an existing chunk it then calls `InMemoryDatasetID.can_read_direct` (in `InMemoryDataset.__getitem__`) which _iterates over all chunks again_

Here's a simple reproducer which writes 100 versions of all zero datasets of increasing size:
```
import shutil

import tempfile

import h5py
import numpy as np
from deshaw.constants.dirs import DIR_cluster_tmp

from versioned_hdf5 import VersionedHDF5File

if __name__ == '__main__':
    # create temp dir path to be cleaned up on exit
    dir = tempfile.mkdtemp(dir=DIR_cluster_tmp())
    path = f'{dir}/data.h5'
    try:
        i = 200
        j = 50_000

        with h5py.File(path, 'w') as f:
            vf = VersionedHDF5File(f)
            with vf.stage_version('r0') as sv:
                sv.create_dataset('values', data=np.zeros((i, j)), chunks=(10, 100))

            for v in range(1, 101):
                with vf.stage_version(f'r{v}') as sv:
                    # every version we grow in both i and j direction
                    new_i = i + 1
                    new_j = j + 500

                    # grow dataset and fill new entries with zeros
                    dataset = sv['values']
                    dataset.resize((new_i, new_j))
                    dataset[i:new_i, :j] = 0
                    dataset[:i, j:new_j] = 0
                    dataset[i:new_i, j:new_j] = 0

                i, j = new_i, new_j
    finally:
        shutil.rmtree(dir)
```
which has the following profiling results (note that I commented out the call to `_verify_new_chunk_reuse`, see https://github.com/deshaw/versioned-hdf5/issues/322, which otherwise makes this even slower...)
```
1708.535 <module>  12655_rewrite_versioned_reproducer_resize.py:1
├─ 1420.831 InMemoryDataset.resize  versioned_hdf5/wrappers.py:608
│  ├─ 1200.752 InMemoryDataset.__getitem__  versioned_hdf5/wrappers.py:653
│  │  ├─ 602.269 InMemoryDatasetID.can_read_direct  versioned_hdf5/wrappers.py:1319
[...]

PR #332, opened 2024-06-06
resize: read the entire Dataset

### LINKED ISSUE #329: Optimize resize further (PyInf#12702)
In #325 we improved the `resize` performance by reading "directly". I tried out the changes and discovered that even reading directly is still pretty slow:
```
978.452 <module>  12655_rewrite_versioned_reproducer_resize.py:1
├─ 689.483 InMemoryDataset.resize  versioned_hdf5/wrappers.py:609
│  ├─ 459.697 InMemoryDataset.get_index  versioned_hdf5/wrappers.py:656
│  │  └─ 459.136 InMemoryDataset.__getitem__  h5py/_hl/dataset.py:749
│  │     └─ 457.641 Reader.read  <built-in>
│  ├─ 150.713 InMemoryDatasetID.data_dict  versioned_hdf5/wrappers.py:1299
│  │  ├─ 92.057 <dictcomp>  versioned_hdf5/wrappers.py:1328
│  │  │  └─ 87.273 spaceid_to_slice  versioned_hdf5/slicetools.py:6
│  │  │     ├─ 49.593 Tuple.__init__  ndindex/ndindex.py:121
│  │  │     │  └─ 47.276 Tuple._typecheck  ndindex/tuple.py:48
│  │  │     │     └─ 28.188 [self]  ndindex/tuple.py
│  │  │     ├─ 21.325 [self]  versioned_hdf5/slicetools.py
│  │  │     └─ 15.047 hyperslab_to_slice  versioned_hdf5/slicetools.py:33
│  │  │        └─ 13.672 Slice.__init__  ndindex/ndindex.py:121
│  │  │           └─ 11.507 Slice._typecheck  ndindex/slice.py:55
│  │  └─ 33.139 <listcomp>  versioned_hdf5/wrappers.py:1318
│  │     └─ 32.294 [self]  versioned_hdf5/wrappers.py
│  ├─ 52.028 ChunkSize.as_subchunks  ndindex/chunking.py:143
│  │  └─ 51.527 _indices  ndindex/chunking.py:260
│  │     ├─ 28.868 <listcomp>  ndindex/chunking.py:263
│  │     │  └─ 23.024 Slice.__init__  ndindex/ndindex.py:121
│  │     │     └─ 14.137 Slice._typecheck  ndindex/slice.py:55
│  │     └─ 19.160 Tuple.__init__  ndindex/ndindex.py:121
│  │        └─ 18.071 Tuple._typecheck  ndindex/tuple.py:48
│  │           └─ 10.284 [self]  ndindex/tuple.py
│  └─ 10.769 [self]  versioned_hdf5/wrappers.py
```
I think this is because even if we can read directly we read each chunk individually (in the `for chunk in self.chunks.as_subchunks(old_shape_idx, size):` loop). I think we could improve this by just doing one read of the entire `Dataset` into an in-memory numpy array and then slicing into that?

ArvidJB:
By reading the entire current Dataset into a numpy array we can improve the `resize` performance even further. Consider the example from #325:
```
import shutil

import tempfile

import h5py
import numpy as np
from deshaw.constants.dirs import DIR_cluster_tmp

if __name__ == '__main__':
    # create temp dir path to be cleaned up on exit
    dir = tempfile.mkdtemp(dir=DIR_cluster_tmp())
    path = f'{dir}/data.h5'
    try:
        i = 200
        j = 50_000

        with h5py.File(path, 'w') as f:
            vf = VersionedHDF5File(f)
                sv.create_dataset('values', data=np.zeros((i, j)), chunks=(10, 100))

            for v in range(1, 101):
                with vf.stage_version(f'r{v}') as sv:
                    # every version we grow in both i and j direction
                    new_i = i + 1
                    new_j = j + 500

                    # grow dataset and fill new entries with zeros
                    dataset = sv['values']
                    dataset.resize((new_i, new_j))
                    dataset[i:new_i, :j] = 0
                    dataset[:i, j:new_j] = 0
                    dataset[i:new_i, j:new_j] = 0

                i, j = new_i, new_j
    finally:
        shutil.rmtree(dir)
```

Prior to #326 it took 1709s. With the optimizations in #326 it takes 978s. With the changes in this commit it now takes only 629s.

Closes #329.

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -638,6 +638,7 @@ class InMemoryDataset(Dataset):
         can_read_direct = self.id.can_read_direct
 
         old_shape_idx = Tuple(*[Slice(0, i) for i in old_shape])
+        old_data = self.get_index(old_shape_idx.raw, can_read_direct=can_read_direct)
         new_data_dict = {}
         for chunk in self.chunks.as_subchunks(old_shape_idx, size):
             if chunk in data_dict:
@@ -645,9 +646,7 @@ class InMemoryDataset(Dataset):
             else:
                 data = np.full(chunk.newshape(size), self.fillvalue, dtype=self.dtype)
                 index = old_shape_idx.as_subindex(chunk)
-                data[index.raw] = self.get_index(
-                    chunk.raw, can_read_direct=can_read_direct
-                )
+                data[index.raw] = old_data[chunk.raw]
                 new_data_dict[chunk] = data
 
         self.id.data_dict = new_data_dict


peytondmurray:
So apparently it is not worth it to read only the individual chunks we need for the resize operation from the dataset, the `h5py` overhead is too high. The one caveat for this is that large datasets will require large amounts of memory. But for smaller datasets this seems worth it. Thoughts?

ArvidJB:
I think the memory overhead is the same: we always read the entire array into memory, either as one array or as many individual chunks, and slicing into the large array should not allocate additional memory.

peytondmurray:
I think we only read the chunks from the underlying dataset which overlap with the new shape. Chunks that remain unchanged just copy their slice mappings to the new dataset:

```python
for chunk in self.chunks.as_subchunks(old_shape_idx, size):
    if chunk in data_dict:
        new_data_dict[chunk] = data_dict[chunk] # <-- just copy over the old slice mapping
    else:
        # In this case we actually need to read the chunk, since it lies at the edge of the new dataset's shape
        data = np.full(chunk.newshape(size), self.fillvalue, dtype=self.dtype)
        index = old_shape_idx.as_subindex(chunk)
        data[index.raw] = self.get_index(
            chunk.raw, can_read_direct=can_read_direct
        )
        new_data_dict[chunk] = data
```

But in any case I think this is probably still a good change to make. Merging now.

ArvidJB:
That's actually a very valid point... If the `data_dict` contains most of the chunks then this is a waste of memory. So we should only read the entire if `can_read_direct`.
Is there a way to make this change without copying the entire code in the for loop?

peytondmurray:
I actually think it may not be possible to read the data in one single operation for the general case. The resize operation takes a rectangular space of the original dataset (solid lines) and drops data outside the new dataset (dashed lines):

```
┌─────────────────────────┐
│                 ┆       │
│                 ┆       │
│                 ┆       │
│   New shape     ┆       │
│                 ┆       │
│                 ┆       │
│                 ┆       │
│-----------------┘       │
│                         │
│      Old shape          │
│                         │
└─────────────────────────┘
```

In the case above, the chunks which intersect the boundary of the new shape aren't a simple rectangular region that can be indexed in one operation, they form an L shape. So you'd need to do at least two indexing operations to pull out the data that needs to be reallocated into new chunks in the resized dataset.

There's probably some arithmetic we can do here by  using `as_subchunks` to get the chunks that intersect the new shape boundary, then we'd have to figure out how they line up with one another.

Alternatively we could construct a boolean index, but that wouldn't be good for really large arrays, which is the problem in the first place. I've asked a few colleagues, and will get back to you on this.

ArvidJB:
Agree, this is trickier than I thought.
Note that we usually call `resize` only when we want to *expand* the array: it's pretty rare to shrink it. Not sure if it helps to consider this as the common case?

Issue #336, opened 2024-06-07
Improve performance of _verify_chunk_reuse (PyInf#11487)


ArvidJB:
The check in `_verify_new_chunk_reuse` is actually fairly slow and dominates commit performance now:
```
1672.012 VersionedHDF5File.stage_version  versioned_hdf5/api.py:267
`- 1671.014 commit_version  versioned_hdf5/versions.py:71
   |- 1280.416 write_dataset  versioned_hdf5/backend.py:114
   |  |- 945.030 _verify_new_chunk_reuse  versioned_hdf5/backend.py:229
   |  |  |- 428.685 Dataset.__getitem__  h5py/_hl/dataset.py:749
   |  |  |  |- 102.522 Dataset.size  h5py/_hl/dataset.py:488
   |  |  |  |  |- 60.019 Dataset.shape  h5py/_hl/dataset.py:467
   |  |  |  |  |  `- 58.882 [self]  h5py/_hl/dataset.py
   |  |  |  |  `- 37.907 prod  <__array_function__ internals>:177
   |  |  |  |- 80.312 [self]  h5py/_hl/dataset.py
   |  |  |  |- 72.637 Dataset._fast_reader  h5py/_hl/dataset.py:527
   |  |  |  |  `- 71.990 [self]  h5py/_hl/dataset.py
   |  |  |  |- 72.460 select  h5py/_hl/selections.py:19
   |  |  |  |  `- 43.554 [self]  h5py/_hl/selections.py
   |  |  |  `- 34.410 Reader.read  <built-in>
   |  |  |- 382.311 inner  contextlib.py:78
   |  |  |  |- 184.283 [self]  contextlib.py
   |  |  |  `- 69.015 _all  numpy/core/_methods.py:61
   |  |  |     |- 34.479 [self]  numpy/core/_methods.py
   |  |  |     `- 34.252 ufunc.reduce  <built-in>
   |  |  `- 122.103 [self]  versioned_hdf5/backend.py
   |  |- 122.336 [self]  versioned_hdf5/backend.py
   |  |- 104.268 Hashtable.hash  versioned_hdf5/hashtable.py:116
   |  |  `- 42.695 openssl_sha256  <built-in>
   |  `- 47.634 Hashtable.__contains__  <frozen _collections_abc>:778
   |     `- 43.601 Hashtable.__getitem__  versioned_hdf5/hashtable.py:205
   |        `- 40.796 [self]  versioned_hdf5/hashtable.py
[...]

PR #337, opened 2024-06-07
Add option to disable verify_chunk_reuse

ArvidJB:
Currently, the verify_chunk_reuse check is relatively slow, see #336. This commit adds an option to disable it by an additional argument to `stage_version(..., verify_chunk_reuse=False)`.

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -179,6 +179,10 @@ def write_dataset(
     slices_to_write = {}
     chunk_size = chunks[0]
 
+    validate_reused_chunks = os.environ.get(
+        "ENABLE_CHUNK_REUSE_VALIDATION", "false"
+    ).lower() in ("1", "true")
+
     with Hashtable(f, name) as hashtable:
         old_chunks = hashtable.largest_index
         chunks_reused = 0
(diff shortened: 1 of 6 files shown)


peytondmurray:
One alternative I could see here is to enable verification through an environment variable. That way the user can set it once for their environment and not have to modify code to get the improved performance you get when no verification is done.

ArvidJB:
I like the environment variable idea: this could be the default if the argument is not specified?

peytondmurray:
Should we merge this as is and then I can add environment variable override support separately?

Also, should the default be to perform validation or to skip validation? I can see motivations for either choice.

ArvidJB:
I think with the environment variable this no longer needs to be exposed on `stage_version`. We can just read the variable directly in `write_dataset`.

peytondmurray:
Okay, chunk verification is now disabled by default, but can be turned on with `ENABLE_CHUNK_REUSE_VALIDATION = 1`. I added a note in the docs about this.

versioned_hdf5/backend.py
```diff
@@ -188,14 +192,17 @@ def write_dataset(
...
-                        raw_dataset=ds,
-                        new_data=data,
-                        data_hash=data_hash,
-                        hashed_slice=hashed_slice,
-                        chunk_being_written=data_s,
-                        slices_to_write=slices_to_write,
-                    )
+                    if os.environ.get(
+                        "ENABLE_CHUNK_REUSE_VALIDATION", "false"
+                    ).lower() in ("1", "true"):
```
ArvidJB, line 197: Can we move this environment variable handling out of the loop?

peytondmurray: 100% - done.

ArvidJB (commented):
LGTM! 👍

PR #347, opened 2024-06-21
Test with numpy 1.24 version now that version 2.* is released

ArvidJB:
Internally we are still on numpy version 1.24. After the numpy 2.0 release that means that we were no longer testing against numpy version 1.* but instead were running tests against numpy 2.* twice.


peytondmurray (approved):
Looks good. Thanks for this!

peytondmurray:
Branch protection rules updated for these new tests.

Issue #331, opened 2024-06-05
Improve exception when attempting to change dataset options (PyInf#12721)


li-dan:
#321 has an example of the exception "ValueError: Compression options can only be specified for the first version of a dataset". It'd be helpful for user debugging to include the name of the dataset, along with the current compression options and the requested compression options. Same for other dataset options that can't be changed, such as chunking, dtype, and fillvalue.

peytondmurray:
Sounds reasonable, although for completeness we'd probably also want to implement some checking of these options in `versioned_hdf5.wrappers._make_new_dset`. I think the most performant way to check these options is to e.g. query `h5py` for the available compression types at module initialization time, then compare the requested compression to the list of available compression types.

li-dan:
Sorry, I mentioned #321, but I'm really concerned about mistakes in application code where `create_dataset` is inadvertently being called with different sets of options, not bugs in h5py. Even a small change in `write_dataset` to enhance the exception messages would be valuable. I want to guard against

li-dan:
Thanks!

PR #358, opened 2024-07-30
Fix nondefault compression handling

### LINKED ISSUE #321: Handle non-default compression (PyInf#12601)
h5py supports a small number of compressions out-of-the-box and also allows you to register additional "compression filters", like the [blosc](https://www.blosc.org/) compressions. The problem is that those are not handled correctly in versioned-hdf5. In this example we are trying to use the "lz4hc" compression:
```
In [1]: import tables  # to register blosc compression

In [2]: with h5py.File('data.h5', 'w') as f:
   ...:         sv.create_dataset('values', data=np.arange(10), compression=32001, compression_opts=(0, 0, 0, 0, 7, 1, 2))
   ...:
[PYFLYBY] from versioned_hdf5 import VersionedHDF5File
[PYFLYBY] import h5py
[PYFLYBY] import numpy as np

In [3]: with h5py.File('data.h5', 'r+') as f:
   ...:     with vf.stage_version('r1') as sv:
   ...:         del sv['values']
   ...:         sv.create_dataset('values', data=np.arange(20), compression=32001, compression_opts=(0, 0, 0, 0, 7, 1, 2))
   ...:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 3
      1 with h5py.File('data.h5', 'r+') as f:
      2     vf = VersionedHDF5File(f)
----> 3     with vf.stage_version('r1') as sv:
      4         del sv['values']
      5         sv.create_dataset('values', data=np.arange(20), compression=32001, compression_opts=(0, 0, 0, 0, 7, 1, 2))

File /opt/python/python-3.11/lib64/python3.11/contextlib.py:144, in _GeneratorContextManager.__exit__(self, typ, value, traceback)
    142 if typ is None:
    143     try:
--> 144         next(self.gen)
    145     except StopIteration:
    146         return False

File /usr/local/python/python-3.11/std/lib64/python3.11/site-packages/versioned_hdf5/api.py:275, in VersionedHDF5File.stage_version(self, version_name, prev_version, make_current, timestamp)
    273     yield group
    274     group.close()
--> 275     commit_version(group, group.datasets(), make_current=make_current,
    276                    chunks=group.chunks,
    277                    compression=group.compression,
    278                    compression_opts=group.compression_opts,
    279                    timestamp=timestamp)
    281     self._log_version_diff_stats(old_current, self.current_version)
    283 except:

File /usr/local/python/python-3.11/std/lib64/python3.11/site-packages/versioned_hdf5/versions.py:132, in commit_version(version_group, datasets, make_current, chunks, compression, compression_opts, timestamp)
    130     slices = write_dataset_chunks(f, name, data.data_dict, shape=data.shape)
    131 else:
--> 132     slices = write_dataset(f, name, data, chunks=chunks[name],
    133                            compression=compression[name],
    134                            compression_opts=compression_opts[name],
    135                            fillvalue=fillvalue)
    136 if shape is None:
    137     if isinstance(data, dict):

File /usr/local/python/python-3.11/std/lib64/python3.11/site-packages/versioned_hdf5/backend.py:109, in write_dataset(f, name, data, chunks, dtype, compression, compression_opts, fillvalue)
    106         raise ValueError("dtype specified but doesn't match already existing dtype")
    108 if compression and compression != ds.compression or compression_opts and compression_opts != ds.compression_opts:
--> 109     raise ValueError("Compression options can only be specified for the first version of a dataset")
    110 if fillvalue is not None and fillvalue != ds.fillvalue:
    111     dtype = ds.dtype

ValueError: Compression options can only be specified for the first version of a dataset
```

The issue here is that h5py returns `None` for all non-standard compressions, see
https://github.com/h5py/h5py/blob/865056bb17a6652ea314391d4f6b4039fab23f2a/h5py/_hl/dataset.py#L574
So we really should check `_filters` instead of `compression` in versioned_hdf5.

peytondmurray:
This PR fixes compression handling for nondefault compression types, closes #321 and partially addresses #331. Additionally if `recreate_dataset` is called, we now log a warning message if compression is enabled for the dataset; this is because `h5py` stores compression settings in `Dataset._filters`, and we have no way of distinguishing these from other filters that may be present in the file.

In practice I think this is an unusual use case, but we now emit a warning if this _does_ happen.

If an invalid compression setting is requested in `write_dataset`, a more informative error message is displayed with a listing of all detected compression settings.


ArvidJB:
Looks good!

Can you add a test for blosc compression via pytables like in https://github.com/deshaw/versioned-hdf5/issues/321#issue-2308777158 ? pytables should be an optional dependency for testing only.

peytondmurray:
I parametrized the tests I added to try with both `hdf5plugin` and `pytables`, but `pytables` appears to be incompatible with `numpy>=2` (https://github.com/PyTables/PyTables/issues/1172). I've made the tests skip if the user has both `pytables` and `numpy>2` installed. Right now if you want to use blosc compression, it seems like `hdf5plugin` is the way to go.

PR #360, opened 2024-08-07
Make the tests workflow update before installing native deps

peytondmurray:
This PR makes the workflow first run `apt update` before attempting to installing native dependencies; currently workflows are broken because `libhdf5` fails to install:

```
Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/c/curl/libcurl4-openssl-dev_7.81.0-1ubuntu1.16_amd64.deb  404  Not Found 
```

This PR should resolve this issue.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,12 +20,8 @@ jobs:
       - name: Install libhdf5-dev
         run: |
           set -xe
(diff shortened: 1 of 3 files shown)


ArvidJB:
Thanks, looks good! I assume that since the checks are passing it works ;-)

PR #362, opened 2024-08-08
Label hypothesis tests as slow

crusaderky:
Follow-up to #345


peytondmurray (approved):
Looks good to me. Thanks for this!

PR #365, opened 2024-08-13
Enforce pre-commit in CI

crusaderky:


peytondmurray (approved):
Thanks for this :rocket:

PR #367, opened 2024-08-13
Sane mypy settings

crusaderky:
Bulk import sane settings for mypy from another project.

Downstream of this, mypy reports 21 errors, all nontrivial. 
Having a fully green mypy run would take substantially more effort and is out of scope for this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,3 +114,18 @@ markers = [
     "setup_args : kwargs for setup fixture.",
     "slow: slow tests",
 ]
+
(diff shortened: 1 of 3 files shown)


peytondmurray:
This seems like a step in the right direction. I haven't added mypy in prior to this because most of this project is untyped and I've been adding type annotations where I can; and because `h5py` itself is untyped. But if we're going to move in the right direction and actually have type checking, we've got to start somewhere. Thanks for this!

PR #370, opened 2024-08-26
[DNM] Redesign InMemoryDataset data model with hyperrectangles selection

crusaderky:
**Note:** This PR is still missing some components - see checklist below.

`versioned_hdf5.wrappers.InMemoryDataset` is plagued by slowness caused by the control code, both because the control code itself is very slow and because it calls `h5py.Dataset.__getitem__` independently for every single chunk.

The new control logic introduced in this PR aims to minimise the number of calls to `h5py.Dataset.__getitem__` by selecting large rectangles of unmodified data spanning multiple chunks.

This new design introduces a helper class, `StagedChangesArray`, which handles all the control logic. This class has minimal external API and is completely abstracted from h5py.

There are also other major design improvements. Namely, in master `__setitem__` first loads all impacted chunks into memory, in the off chance that they are not wholly covered by the parameter index, and then updates them. In this PR, this happens exclusively for the chunks that are not wholly covered by the index of `__setitem__`.
In other words, `a[:] = 1` in master loads the whole dataset into memory if it's not in memory already, whereas in this PR it never touches the disk.

Analogously, in master `resize()` reads the whole dataset from disk. This no longer happens in this PR, which only loads the edge chunks if the shape is not exactly divisible by chunk_size.

## Status
- [x] implement `__getitem__`
- [x] implement `__setitem__`
- [x] implement `resize()`
- [x] crudely test `__getitem__` by hand
- [x] crudely test `__setitem__` by hand
- [x] crudely test `resize()` by hand
- [x] integrate with wrappers.py
- [x] all existing unit tests pass
- [ ] new unit tests to stress the new classes in isolation
- [ ] clean up legacy code
- [x] high level documentation
- [ ] break PR up into stages to simplify code review

## Demo
See `demo.ipynb`.

## Benchmarks
See `benchmarks.ipynb`.
Early benchmarks show up to 500x speedup in the control code and up to 2000x reduction in the number of calls to `h5py.Dataset.__getitem__`.

PR #371, opened 2024-09-10
[InMemoryDataset redesign] fill_hyperspace()

crusaderky:
This is a partial PR that breaks down #370 into digestible chunks - see https://github.com/deshaw/versioned-hdf5/pull/370#issuecomment-2341012896

This PR adds a single new utility function, `fill_hyperspace`, that for the time being remains unused and will be later used by #370.


versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,355 @@
...
+import numpy as np
+from cython import Py_ssize_t
+
+_empty_tpl = array.array("l", [])
+_py_ssize_t_nbytes = _empty_tpl.itemsize
+# assert that "long" is the same length as void* on the current platform
+# Note: Py_ssize_t is always the same as np.intp
+assert np.intp().nbytes == _py_ssize_t_nbytes
+
+if cython.compiled:  # pragma: nocover
```
crusaderky, line 18: One of the many things I love about cython pure python mode: you can just disable cythonization and run coverage on it!

versioned_hdf5/hyperspace.pyx
```diff
@@ -0,0 +1 @@
+hyperspace.py
```
crusaderky, line 1: This is a symlink hack to make meson compile a pure-python cython module

versioned_hdf5/meson.build
```diff
@@ -26,6 +32,38 @@ py.extension_module(
...
+    # Note: this is possible only because we went through every single // and % in
+    # the module and made sure the sign of the two operands is always the same.
+    # Comment this out if behaviour changes when not compiled.
+    '-X', 'cdivision=True',
+]
+
+# To generate HTML compilation reports:
+# (First double check that flags below match those above)
+#
+# cythonize -3 -a -X infer_types=True -X initializedcheck=False -X boundscheck=False -X wraparound=False -X cdivision=True versioned_hdf5/{hyperspace,subchunk_map}.py
```
crusaderky, line 53: This line is incorrect until the next PR that will add the same flags to subchunk_map.py

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,367 @@
...
+    from cython.cimports import numpy as np  # type: ignore
```
crusaderky, line 19: `cimport`'ing numpy is a nice to have. All it does is produce slightly neater and shorter C code whenever numpy is invoked.

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,366 @@
...
+            shape[1:],
+            out_a[:, 1:],
+            out_b[:, 1:],
+            cur,
+        )
+        prefix = obstacle1[0]
+        out_a[cur:new_cur, 0] = prefix
+        out_b[cur:new_cur, 0] = prefix + 1
+        return new_cur
+
```
crusaderky, line 254: What comes below cost me two solid days of headache trying to visualize in my head spaces with more than 3 dimensions

versioned_hdf5/subchunk_map.py
```diff
@@ -17,6 +17,13 @@
 )
 from numpy.typing import NDArray
 
+# Temporary hack to work around pytest issue: if a pure-pyton Cython module is first
+# imported by `pytest .` and it contains doctests, it will be imported in pure python
+# instead of its compiled form. This fails in CI, as Cython is not a runtime dependency,
+# and regardless would cause a lot of type-checking normally performed by Cython to be
+# skipped so it is to be avoided.
+from . import hyperspace  # noqa: F401
```
crusaderky, line 25: This will later be replaced by a proper import of functionality

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,363 @@
...
+) -> Py_ssize_t[:, :]:
+    """Given a N-dimensional space of the given shape and a series of obstacles each one
+    point in size, generate hyperrectangles that cover the whole remaining space
+    without ever crossing an obstacle or another hyperrectangle.
+
+    Parameters
+    ----------
+    obstacles:
+        A list of points to avoid, for example as returned by
+        versioned_hdf5.subchunk_map._modified_chunks_in_selection
```
peytondmurray, line 68: One thing to note here unless I missed it somewhere else - I think these have to be ordered along each dimension, right? This is noted in `_hyperrectangles_between`, but it might be good to include here too.

versioned_hdf5/subchunk_map.py
peytondmurray: Understood. Maybe the later revision will address this, but 

1. we can also make `cython` a test dependency if needed, and
2. I don't think we have any doctests?

peytondmurray (approved):
This first addition looks great. Thank you especially for the very thorough documentation, because `_hyperrectangles_between` would be quite hard to follow without it! :rocket:

versioned_hdf5/subchunk_map.py
crusaderky: hyperspace.py has two doctests.
Making cython a test dependency would be a bad idea because, while it would fix the CI failure, it would cause ALL tests to use the pure-python version of hyperspace.py, which is much more lenient in terms of typing than the cythonized version, thus potentially hiding breaking bugs completely until someone manually tries using it.

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,363 @@
...
+from cython import Py_ssize_t
+
+_empty_tpl = array.array("l", [])
+_py_ssize_t_nbytes = _empty_tpl.itemsize
+# assert that "long" is the same length as void* on the current platform
+# Note: Py_ssize_t is always the same as np.intp
+assert np.intp().nbytes == _py_ssize_t_nbytes
+
+if cython.compiled:  # pragma: nocover
+    from cython.cimports.cpython import array  # type: ignore
```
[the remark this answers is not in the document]
crusaderky: I don't follow, what types?

versioned_hdf5/hyperspace.py
crusaderky: Yes. Amended.

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,363 @@
...
+    FIXME
+    -----
+    The current implementation performs suboptimally in the common use case of an
+    uninterrupted sequence of obstacles along an axis that is not the innermost one:
+
+        Input     Current output  Optimal output
+      0123456789     0123456789      0123456789
+    0 .X........   0 aXbbbbbbbb    0 aXbbbbbbbb
+    1 .X........   1 cXdddddddd    1 aXbbbbbbbb
+    2 .X........   2 eXffffffff    2 aXbbbbbbbb
```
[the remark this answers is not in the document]
crusaderky: Done: https://github.com/deshaw/versioned-hdf5/issues/375

versioned_hdf5/hyperspace.py
```diff
@@ -0,0 +1,363 @@
...
+_py_ssize_t_nbytes = _empty_tpl.itemsize
+# assert that "long" is the same length as void* on the current platform
+# Note: Py_ssize_t is always the same as np.intp
+assert np.intp().nbytes == _py_ssize_t_nbytes
+
+if cython.compiled:  # pragma: nocover
+    from cython.cimports.cpython import array  # type: ignore
+
+    @cython.ccall
+    @cython.inline
```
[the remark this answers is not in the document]
crusaderky: Zero. Not even in a sythetic test that does nothing but call the function 1 million times in a row. I'm removing it.

PR #372, opened 2024-09-10
[InMemoryDataset redesign] Object-oriented as_subchunk_map

crusaderky:
- Part of #386

Refactor `as_chunk_submap` to be built on top of an abstract class, `IndexChunkMapper`, which is then implemented by one subclass per type of index.
This in and by itself is hardly useful, but it will become necessary when these mapper objects will be used all over the place by #370.

Additionally:

- Move the data type for all indices in the module from PySsize_t to hsize_t, to make them compatible with libhdf5 and `slicetools.pyx::read_many_slices()`. This integration will materialize only in a later PR.
- Sanitize arguments order and naming conventions to make it coherent with later additions

### DIFF
--- a/versioned_hdf5/meson.build
+++ b/versioned_hdf5/meson.build
@@ -78,4 +78,5 @@ py.extension_module(
     install: true,
     subdir: 'versioned_hdf5',
     dependencies: compiled_deps,
+    cython_args: cython_args,
 )
--- a/versioned_hdf5/subchunk_map.py
+++ b/versioned_hdf5/subchunk_map.py
@@ -200,28 +200,6 @@ class IntegerMapper(IndexChunkMapper):
 
 
 @cython.cclass
-class EverythingMapper(IndexChunkMapper):
-    """Select all points along an axis [:].
-
-    This is functionally identical to SliceMapper(slice(None), chunk_size, dset_size),
-    special-cased here for simplicity and speed.
-    """
-
-    def __init__(self, dset_size: hsize_t, chunk_size: hsize_t):
-        n_chunks = ceil_a_over_b(dset_size, chunk_size)
-        super().__init__(np.arange(n_chunks, dtype=np_hsize_t), dset_size, chunk_size)
-
-    @cython.ccall
-    def chunk_submap(self, chunk_idx: hsize_t) -> tuple[Slice, slice, slice]:
-        chunk_start, chunk_stop = self._chunk_start_stop(chunk_idx)
-        return (
-            Slice(chunk_start, chunk_stop, 1),
-            slice(chunk_start, chunk_stop, 1),
-            slice(0, chunk_stop - chunk_start, 1),
-        )
-
-
-@cython.cclass
 class IntegerArrayMapper(IndexChunkMapper):
     """IndexChunkMapper for one-dimensional fancy integer array indices.
     This is also used for boolean indices (preprocessed with np.flatnonzero()).
@@ -302,7 +280,7 @@ def index_chunk_mappers(
     # Handle the remaining suffix axes on which we did not select, we still need to
     # break them up into chunks.
     for d, n in zip(shape[idx_len:], chunk_size[idx_len:]):
-        mappers.append(EverythingMapper(d, n))
+        mappers.append(SliceMapper(slice(0, d, 1), d, n))
 
     return idx, mappers
 
@@ -322,10 +300,7 @@ def _index_to_mapper(idx, dset_size: hsize_t, chunk_size: hsize_t) -> IndexChunk
         return IntegerArrayMapper(idx, dset_size, chunk_size)
 
     if isinstance(idx, slice):
-        if idx == slice(0, dset_size, 1):
-            return EverythingMapper(dset_size, chunk_size)
-        else:
-            return SliceMapper(idx, dset_size, chunk_size)
+        return SliceMapper(idx, dset_size, chunk_size)
 
     raise NotImplementedError(f"index type {type(idx)} not supported")
 
--- a/versioned_hdf5/subchunk_map.py
+++ b/versioned_hdf5/subchunk_map.py
@@ -1,32 +1,41 @@
+# Note: this entire module is compiled by cython with wraparound=False
+# See meson.build for details
+
 from __future__ import annotations
 
+import abc
+import enum
 import itertools
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterator
 from typing import TYPE_CHECKING, Any
 
 import cython
 import numpy as np
-from cython import Py_ssize_t
-from ndindex import (
-    BooleanArray,
-    ChunkSize,
-    Integer,
-    IntegerArray,
-    Slice,
-    Tuple,
-    ndindex,
-)
+
+# Use same data type for indexing as in libhdf5 C.
+# This matters on 32-bit platforms, where ssize_t is 32 bit and would be incapable of
+# indexing hdf5 datasets on disk wider than 2**31
+from cython import bint
+from ndindex import ChunkSize, Slice, Tuple, ndindex
 from numpy.typing import NDArray
 
-from .cytools import ceil_a_over_b, smallest_step_after
+from .cytools import ceil_a_over_b, count2stop, np_hsize_t, smallest_step_after
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: nocover
     # TODO import from typing and remove quotes (requires Python 3.10)
     # TODO use type <name> = ... (requires Python 3.12)
     from typing_extensions import TypeAlias
 
-    AnySlicer: TypeAlias = (
-        "slice | NDArray[np.intp] | NDArray[np.bool] | tuple[()] | Py_ssize_t"
+    AnySlicer: TypeAlias = "slice | NDArray[np_hsize_t] | int"
+    AnySlicerND: TypeAlias = tuple[AnySlicer, ...]
+
+if cython.compiled:  # pragma: nocover
+    from cython.cimports.versioned_hdf5.cytools import (  # type: ignore
+        ceil_a_over_b,
+        count2stop,
+        hsize_t,
+        smallest_step_after,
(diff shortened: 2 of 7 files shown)


versioned_hdf5/meson.build
```diff
@@ -16,6 +16,12 @@ compiled_deps = [
     dependency('mpi'),
 ]
 
+# From https://numpy.org/devdocs//reference/random/examples/cython/meson.build.html
+npy_root = run_command(py, ['-c',
+    'import os; os.chdir(".."); import numpy; print(os.path.abspath(numpy.get_include() + "../../.."))'
+    ], check: true).stdout().strip()
+npy_include_path = npy_root / '_core' / 'include'
+
```
crusaderky, line 24: This is all the same as in #371

versioned_hdf5/subchunk_map.py
```diff
@@ -17,308 +22,678 @@
 )
 from numpy.typing import NDArray
 
+if cython.compiled:
+    from cython.cimports import numpy as np  # type: ignore
+
+    np.import_array()  # type: ignore[attr-defined]
```
crusaderky, line 28: See https://github.com/deshaw/versioned-hdf5/pull/371#discussion_r1752066252

versioned_hdf5/subchunk_map.py
```diff
@@ -17,308 +22,678 @@
...
+@cython.exceptval(check=False)
 def _ceiling(a: Py_ssize_t, b: Py_ssize_t) -> Py_ssize_t:
-    """Returns ceil(a/b)"""
-    return -(-a // b)
+    """Returns ceil(a/b). Assumes a >= 0 and b > 0."""
+    # Note: this module is compiled with the cython.cdivision flag.
+    # This causes behaviour to change if a and b have opposite signs and you try
+    # debugging the module in pure python, without compiling it.
+    # This function blindly assumes that a and b are always the same sign.
+    return a // b + (a % b > 0)
```
crusaderky, line 48: No performance difference observed

versioned_hdf5/subchunk_map.py
```diff
@@ -17,308 +22,678 @@
...
+        shift: bint,
+    ) -> tuple[int, int]:
+        # Raise if we're selecting an index out of bounds
+        start, _ = self.chunk_start_stop(chunk_start_idx, chunk_stop_idx)
+        return self.idx, self.idx - start if shift else self.idx
+
+
+@cython.cclass
+class EverythingMapper(IndexChunkMapper):
+    """Select all points along an axis [:]"""
```
crusaderky, line 548: Tecnically the same as `SliceMapper(0, dset_size, 1, chunk_size, dset_size)` but much cleaner and faster

versioned_hdf5/meson.build
```diff
@@ -26,6 +26,29 @@ py.extension_module(
...
+]
+
+# To generate HTML compilation reports:
+# (First double check that flags below match those above)
+#
+# cythonize -3 -a -X infer_types=True -X initializedcheck=False -X boundscheck=False -X wraparound=False -X cdivision=True versioned_hdf5/{hyperspace,subchunk_map}.py
+#
+# Or uncomment the -a flag above and then run:
+# meson build && pushd build && meson compile && popd && firefox $(find build/ -name "*.html")
+
```
crusaderky, line 51: All of the above is a copy-paste from #371 - replicated here to allow for parallel review

versioned_hdf5/tests/test_subchunk_map.py
```diff
@@ -10,16 +10,22 @@
...
 
 
 def non_negative_step_slices(size: int):
-    start = st.one_of(st.integers(-size - 1, size + 1), st.none())
-    stop = st.one_of(st.integers(-size - 1, size + 1), st.none())
+    start = st.one_of(st.none(), st.integers(-size - 1, size + 1))
+    stop = st.one_of(st.none(), st.integers(-size - 1, size + 1))
     # only non-negative steps (or None) are allowed
-    step = st.one_of(st.integers(1, size + 1), st.none())
+    step = st.one_of(st.none(), st.integers(1, size + 1))
```
crusaderky, line 28: This is a nit following hypothesis best practices from https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.one_of

In practice this doesn't matter due to the very high cardinality of our tests and the fact that all slices are reduced by ndindex very early on.

versioned_hdf5/subchunk_map.py
```diff
@@ -36,280 +52,405 @@
...
-    __getitem__: out[a] = cache_chunk[b]
-    __setitem__: cache_chunk[b] = value[a]
-    """
-    start: Py_ssize_t = max(c_start, s_start)
-    # Get the smallest lcm multiple of common that is >= start
-    start = smallest_step_after(start, s_start % s_step, s_step)
-    # Finally, we need to shift start so that it is relative to index
-    start = (start - s_start) // s_step
+class DropAxis(enum.Enum):
+    _drop_axis = 0
```
crusaderky, line 56: replaces `()` for scalar indices, for the sake of code readability. It took me a long time to realise that the `()` is never actually used to index an ndarray as it is discarded at the end of `as_subchunk_map`.

versioned_hdf5/subchunk_map.py
```diff
@@ -36,280 +52,405 @@
...
 
-    stop: Py_ssize_t = min(c_stop, s_stop)
-    stop = ceil_a_over_b(stop - s_start, s_step) if stop > s_start else 0
 
-    return slice(int(start), int(stop), 1)
+# Returned instead of an AnySlicer. Signals that the axis should be removed when
+# aggregated into an AnySlicerND.
+DROP_AXIS = DropAxis._drop_axis
```
crusaderky, line 61: replaces `()` for scalar indices, for the sake of code readability. It took me a long time to realise that the `()` is never actually used to index an ndarray as it is discarded at the end of `as_subchunk_map`.

versioned_hdf5/subchunk_map.py
```diff
@@ -36,280 +52,405 @@
...
+        self,
+        chunk_indices: hsize_t[:],
+        dset_size: hsize_t,
+        chunk_size: hsize_t,
+    ):
+        self.chunk_indices = chunk_indices
+        self.dset_size = dset_size
+        self.chunk_size = chunk_size
+        self.n_chunks = ceil_a_over_b(dset_size, chunk_size)
+        self.last_chunk_size = (dset_size % chunk_size) or chunk_size
```
crusaderky, line 95: Unused for now

versioned_hdf5/subchunk_map.py
```diff
@@ -36,280 +52,405 @@
...
-    See examples on _subindex_chunk_slice
+@cython.cclass
+class IndexChunkMapper:
+    """Abstract class that manipulates a numpy fancy index along a single axis of a
+    chunked array
+
+    Parameters
+    ----------
+    chunk_indices:
+        Array of indices of all the chunks involved in the selection along the axis
```
crusaderky, line 72: For the purpose of `as_subchunk_map` exclusively, it would have been more efficient writing a function that returns an iterator of chunk indices. However, later PRs extensively rely on this being an actual array.

versioned_hdf5/subchunk_map.py
```diff
@@ -36,280 +52,405 @@
...
+    if idx_len == 2:
+        return slice(int(start), int(stop), int(step))
+
+    if stop2count(start, stop, step) == idx_len:
+        j = start + step + step
+        for i in range(2, idx_len):
+            if idx_v[i] != j:
+                break
+            j += step
+        else:
```
crusaderky, line 397: functionally equivalent to `if (idx == np.arange(start, stop, step)).all():`, but faster

versioned_hdf5/subchunk_map.py
```diff
@@ -36,294 +45,353 @@
...
+        """Return the range of points [a, b[ of the chunk indexed by chunk_idx"""
+        if not cython.compiled:
+            chunk_idx = int(chunk_idx)  # np.uint64 + int -> np.float64
+
+        start = chunk_idx * self.chunk_size
+        stop = min(start + self.chunk_size, self.dset_size)
+        return start, stop
+
+    @cython.ccall
+    @abc.abstractmethod
```
peytondmurray, line 99: Is it a problem that `IndexChunkMapper` doesn't currently inherit from `abc.ABCMeta` when using [this decorator](https://docs.python.org/3/library/abc.html#abc.abstractmethod)?

peytondmurray (approved):
Looks good - I didn't go through the math line by line, but its correctness should be ensured by the growing body of tests.

versioned_hdf5/subchunk_map.py
crusaderky: It is, actually. It means that if you forget to override an abstractmethod nothing bad happens and you end up with a dummy method that returns None.
However, cython breaks if you try either `class IndexChunkMapper(abc.ABC)` or `class IndexChunkMapper(metaclass=abc.ABCMeta)`. So the `abc.abstractmethod` decorator remains purely descriptive. 

*sigh*.
I'll write a note about it.

versioned_hdf5/subchunk_map.py
```diff
@@ -36,294 +45,353 @@
...
+        self.dset_size = dset_size
+        self.chunk_size = chunk_size
+
+    @cython.cfunc
+    @cython.nogil
+    @cython.exceptval(check=False)
+    def _chunk_start_stop(self, chunk_idx: hsize_t) -> tuple[hsize_t, hsize_t]:
+        """Return the range of points [a, b[ of the chunk indexed by chunk_idx"""
+        if not cython.compiled:
+            chunk_idx = int(chunk_idx)  # np.uint64 + int -> np.float64
```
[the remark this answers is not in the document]
crusaderky: Less of "nice catch" and more of "nice 2 hours staring at the monitor in disbelief trying to figure out why everything falls apart as soon as you stop compiling" :facepalm:

PR #374, opened 2024-09-11
[InMemoryDataset redesign] Overhaul test_subchunk_map

crusaderky:
Rework test_subchunk_map:
- Remove dependency from ndindex. Don't assume that ndindex does the right thing.
- Rework the hypothesis strategy so that it's separate from the test. This is propaedeutic to reuse the same strategy later on in different tests.
- Add use case of scalar indices (it was completely missing!)
- Add use case of slice start, stop, and/or step exceeding array size
- Add use case of index shorter than ndim
- Add use case of chunk size = 1
- Add use case of integer array indices with repeated elements, possibly larger than original array

### DIFF
--- a/versioned_hdf5/tests/test_subchunk_map.py
+++ b/versioned_hdf5/tests/test_subchunk_map.py
@@ -1,136 +1,135 @@
-from collections import defaultdict
+from __future__ import annotations
+
+from typing import Any
 
 import hypothesis
-import ndindex
 import numpy as np
 import pytest
 from hypothesis import given
 from hypothesis import strategies as st
 from hypothesis.extra import numpy as stnp
-from numpy.testing import assert_equal
+from numpy.testing import assert_array_equal
 
 from ..subchunk_map import as_subchunk_map
 
+max_examples = 10_000
 
-def non_negative_step_slices(size):
-    start = st.one_of(st.integers(min_value=-size, max_value=size - 1), st.none())
-    stop = st.one_of(st.integers(min_value=-size, max_value=size), st.none())
+
+def non_negative_step_slices(size: int):
+    start = st.one_of(st.integers(-size - 1, size + 1), st.none())
+    stop = st.one_of(st.integers(-size - 1, size + 1), st.none())
     # only non-negative steps (or None) are allowed
-    step = st.one_of(st.integers(min_value=1, max_value=size), st.none())
+    step = st.one_of(st.integers(1, size + 1), st.none())
     return st.builds(slice, start, stop, step)
 
 
-@pytest.mark.slow
-@given(st.data())
-@hypothesis.settings(database=None, max_examples=10_000, deadline=None)
-def test_as_subchunk_map(data):
-    ndim = data.draw(st.integers(1, 4), label="ndim")
-    shape = data.draw(st.tuples(*[st.integers(1, 100)] * ndim), label="shape")
-    chunks = data.draw(st.tuples(*[st.integers(5, 20)] * ndim), label="chunks")
-    idx = ndindex.Tuple(
-        *[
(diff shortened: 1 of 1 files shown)


versioned_hdf5/tests/test_subchunk_map.py
```diff
@@ -1,136 +1,135 @@
...
 
 import hypothesis
-import ndindex
 import numpy as np
 import pytest
 from hypothesis import given
 from hypothesis import strategies as st
 from hypothesis.extra import numpy as stnp
-from numpy.testing import assert_equal
+from numpy.testing import assert_array_equal
```
peytondmurray, line 11: As I recall there is some difference here with `assert_equal`, in that `assert_equal` is defines special handling for non-array types and is thus a little more strict about what it considers equal. Are there any implications of changing to the more lenient `assert_array_equal`?

versioned_hdf5/tests/test_subchunk_map.py
```diff
@@ -1,136 +1,135 @@
...
 
+max_examples = 10_000
+
 
-def non_negative_step_slices(size):
-    start = st.one_of(st.integers(min_value=-size, max_value=size - 1), st.none())
-    stop = st.one_of(st.integers(min_value=-size, max_value=size), st.none())
+def non_negative_step_slices(size: int):
+    start = st.one_of(st.integers(-size - 1, size + 1), st.none())
+    stop = st.one_of(st.integers(-size - 1, size + 1), st.none())
```
peytondmurray, line 20: Since you're defining slices, the `max_value` can be anything and still work; but for testing purposes why is `max_value` not `size` rather than `size+1`?

versioned_hdf5/tests/test_subchunk_map.py
```diff
@@ -1,136 +1,135 @@
...
-        ]
+@st.composite
+def basic_idx_st(draw, shape: tuple[int, ...]) -> Any:
+    """Hypothesis draw of slice and integer indexes"""
+    nidx = draw(st.integers(0, len(shape)))
+    idx_st = st.tuples(
+        *(
+            # FIXME we should push the scalar use case into non_negative_step_slices
+            # However ndindex fails when mixing scalars and slices and array indices
+            # https://github.com/Quansight-Labs/ndindex/issues/188
```
peytondmurray, line 34: Shall we make an issue so that we can track this work?

peytondmurray (approved):
Thank you for this - just a few other questions, but I think this looks good otherwise :rocket:

versioned_hdf5/tests/test_subchunk_map.py
crusaderky: From https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html:

> When either actual or desired is already an instance of [numpy.ndarray](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray) and desired is not a dict, the behavior of assert_equal(actual, desired) is identical to the behavior of assert_array_equal

The main difference is that in older versions of numpy assert_array_equal supports strict=True, while it's only been added later to assert_equal.

versioned_hdf5/tests/test_subchunk_map.py
crusaderky: Because having a slice that exceeds the size of the array being sliced is an edge use case worth testing. Now, in practice, this edge case is negated by the call to `ndindex.ndindex.reduce()` which happens inside as_chunk_submap, but I don't want to assume it will always be there in this test.

versioned_hdf5/tests/test_subchunk_map.py
crusaderky: on  https://github.com/Quansight-Labs/ndindex/issues/188 @asmeurer and I reached consensus that this is not worth pursuing

PR #378, opened 2024-10-02
[InMemoryDataset redesign] Read many slices at once with the HDF5 C API

crusaderky:
This is an alternative to #370 after it's been found out that virtual datasets perform very poorly in libhdf5.

Add a function to quickly read potentially thousands of slices from HDF5 into a numpy array, or between numpy arrays. This new function remains dormant for now and will be used in a later PR by a variant of StagedChangesArray from #370.

### DIFF
--- a/versioned_hdf5/tests/test_slicetools.py
+++ b/versioned_hdf5/tests/test_slicetools.py
@@ -1,8 +1,16 @@
+from typing import Any
+
+import hypothesis
 import numpy as np
+import pytest
 from h5py._hl.selections import Selection
+from hypothesis import given
+from hypothesis import strategies as st
 from numpy.testing import assert_equal
 
-from ..slicetools import spaceid_to_slice
+from ..slicetools import count2stop, read_many_slices, spaceid_to_slice, stop2count
+
+max_examples = 10_000
 
 
 def test_spaceid_to_slice(h5file):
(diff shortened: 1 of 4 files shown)


versioned_hdf5/tests/test_slicetools.py
```diff
@@ -36,3 +44,400 @@ def test_spaceid_to_slice(h5file):
...
+            src, dst[::2, ::3], [(0, 0), (2, 0)], [(0, 0), (2, 0)], (2, 4), fast=True
+        )
+
+    # For the sake of awareness, user must explicitly pass fast=False
+    with pytest.raises(NotImplementedError, match="contiguous"):
+        read_many_slices(src, dst[::2, ::3], [(0, 0), (2, 0)], [(0, 0), (2, 0)], (2, 4))
+
+
+def test_read_many_slices_not_fast_read_ok(h5file):
+    """src is a h5py dataset that doesn't support fast read"""
```
peytondmurray, line 373: It might be good to outline somewhere in the docs when a dataset supports fast reading, since there is a significant difference in performance.

peytondmurray (approved):
Thanks for this - the tests seem comprehensive, and I think there's only one place where we might need a lock. Otherwise this looks great! :rocket:

PR #381, opened 2024-10-15
Repair pre-commit CI

crusaderky:
The pre-commit CI job has recently started misbehaving due to the upgrade of ubuntu-latest from 22.04 to 24.04.
e.g. https://github.com/deshaw/versioned-hdf5/actions/runs/11348180310/job/31561463192?pr=380

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,9 +9,10 @@ on:
 jobs:
   checks:
     name: pre-commit hooks
-    # runs-on: ubuntu-latest  # ubuntu-24.04 breaks pre-commit/action@v3.0.1
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4.2.1
       - uses: actions/setup-python@v5.2.0
+        with:
+          python-version: '3.12' 
       - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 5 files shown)


peytondmurray:
Saw this last night as well. It sounds like for some reason switching to 24.04 causes `actions/setup-python` to _not_ install python if one is already found on $PATH (???), which I guess was not the case for 22.04 (but is for 24.04). Instead of pinning to an older image, I think moving forward we will need to explicitly request for

```
with:
  python-version: 3.x
```

as seen here: https://github.com/pre-commit/action/issues/210

peytondmurray (commented):
Saw this last night as well. It sounds like for some reason switching to 24.04 causes actions/setup-python to not install python if one is already found on $PATH (???), which I guess was not the case for 22.04 (but is for 24.04). Instead of pinning to an older image, I think moving forward we will need to explicitly request for

with:
  python-version: 3.x

as seen here: https://github.com/pre-commit/action/issues/210. Thoughts?

crusaderky:
> Saw this last night as well. It sounds like for some reason switching to 24.04 causes actions/setup-python to not install python if one is already found on $PATH (???), which I guess was not the case for 22.04 (but is for 24.04). Instead of pinning to an older image, I think moving forward we will need to explicitly request for
> 
> with: python-version: 3.x
> 
> as seen here: [pre-commit/action#210](https://github.com/pre-commit/action/issues/210). Thoughts?

Yep, this works. Fixed!

peytondmurray:
Awesome, thank you!

PR #382, opened 2024-10-15
Centralized cython tools; redefine hsize_t

crusaderky:
- Refactor some general purpose utility functions into their own module. This will become more important in later PRs as it turns out it's exceptionally hard to cimport anything from slicetools.pyx.
- Review the definition of hsize_t and align it with libhdf5. See comment in the code for full discussion.

### DIFF
--- a/versioned_hdf5/cytools.pxd
+++ b/versioned_hdf5/cytools.pxd
@@ -1,0 +1,38 @@
+# This file allows cimport'ing the functions and types declared below from other
+# Cython modules
+
+from cython cimport Py_ssize_t
+from libc.stdint cimport uint64_t
+
+# Centralized definition of hsize_t, in accordance with libhd5:
+# https://github.com/HDFGroup/hdf5/blob/6b43197b0817596f47670c6b55d26ff7f86d6bd9/src/H5public.h#L301
+#
+# versioned_hdf5 uses the same datatype for indexing as libhdf5. Notably, this differs
+# from numpy's npy_intp (same as Py_ssize_t, ssize_t, and signed long), which is only 32
+# bit wide on 32 bit platforms, to allow indexing datasets with >=2**31 points per axis
+# on disk, as long as you don't load them in memory all at once.
+#
+# Note that hsize_t is unsigned, which can lead to integer underflows.
+#
+# The definition of hsize_t has changed over time in the libhdf5 headers, as noted
+# below.
+# C99 dictates that long long is *at least* 64 bit wide, while uint64_t is *exactly* 64
+# bit wide. So the two definitions are de facto equivalent.
+# The below definition is inconsequential, as it is overridden by whatever version of
+# hdf5.h is installed thanks to the 'cdef extern from "hdf5.h"' block.
+#
+# h5py gets this wildly wrong, as it defines hsize_t as long long, which is signed, and
+# confusingly also defines hssize_t as signed long long - which is an alias for
+# long long:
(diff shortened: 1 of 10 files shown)


versioned_hdf5/cytools.pxd
```diff
@@ -0,0 +1,32 @@
...
+from cython cimport Py_ssize_t
+from libc.stdint cimport uint64_t as hsize_t
+
+
+cpdef hsize_t stop2count(hsize_t start, hsize_t stop, hsize_t step) noexcept nogil
+cpdef hsize_t count2stop(hsize_t start, hsize_t count, hsize_t step) noexcept nogil
+cpdef Py_ssize_t ceil_a_over_b(Py_ssize_t a, Py_ssize_t b) noexcept nogil
+cpdef Py_ssize_t smallest_step_after(
+    Py_ssize_t x, Py_ssize_t a, Py_ssize_t m
+) noexcept nogil
```
crusaderky, line 32: These two functions will be changed to hsize_t in a later PR

versioned_hdf5/slicetools.pyx
```diff
@@ -1,3 +1,4 @@
+# distutils: language=c++
```
crusaderky, line 1: Moved from meson.build

versioned_hdf5/slicetools.pyx
```diff
@@ -32,10 +36,6 @@ cdef extern from "hdf5.h":
     ctypedef int hbool_t
     ctypedef int herr_t
     ctypedef int htri_t
-    ctypedef long long hsize_t  # as per C99, uint64 or wider
```
crusaderky, line 35: This was actually signed!

versioned_hdf5/tests/test_cytools.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@given(st.data(), st.integers(0, 10), st.integers(1, 4))
+def test_smallest_step_after(data, x, m):
+    a = data.draw(st.integers(0, x))
+    expect = a
+    while expect < x:
+        expect += m
+    actual = smallest_step_after(x, a, m)
+    assert actual == expect
```
crusaderky, line 43: New unit tests

versioned_hdf5/meson.build
```diff
@@ -61,11 +60,21 @@ cython_args = [
 # meson build && pushd build && meson compile && popd && firefox $(find build/ -name "*.html")
 
 py.extension_module(
-    'subchunk_map',
+    'cytools',
```
peytondmurray, line 63: I see the comment below - in the future it might be possible to do this with Meson generators, I think? Anyway, this is fine for now.

versioned_hdf5/cytools.pxd
peytondmurray: I guess later on when this becomes `hsize_t` we can pin libhdf5 to the appropriate version in `meson.build`?

peytondmurray (approved):
A few small comments, otherwise this looks good :rocket:

versioned_hdf5/cytools.pxd
crusaderky: I found out how to `#include "hdf5.h"` in cython: `cdef extern from "hdf5.h"` causes cython to completely disregard whatever you're writing below (yes, really :facepalm: ) and import from the C header instead

versioned_hdf5/meson.build
crusaderky: I'm hesitant to call `ln -s cytools.py cytools.pyx` at build time, as it would break Windows builds, whereas symlink checked in git are a very well known problem for windows users that's easily solved with a stackoverflow search.

PR #384, opened 2024-10-17
[InMemoryDataset redesign] New function asarray()

crusaderky:
Part of #386 

Variant of np.asarray(a, dtype=dtype), with two differences:
1. If a is a numpy-like array, don't coerce it to a numpy.ndarray
2. If a has a ABI-compatible dtype, return a view instead of a copy

Works around https://github.com/numpy/numpy/issues/27509


peytondmurray (approved):
Looks like a nice opportunity to grab views when possible :rocket:

PR #385, opened 2024-10-17
[InMemoryDataset redesign] build_indices_and_offsets

crusaderky:
Part of #386 

Rewrite `build_data_dict` to serve the new StagedChangesArray introduced by #386.
This PR adds a new function, which remains dormant side-by-side to `build_data_dict`, and will eventually replace it.

## Testing
Besides unit tests, I've temporarily rigged running this function every time `build_data_dict` runs (see reverted commit) and never crashed. Definitive proof that this function returns correct results throughout the whole test suite is that #386, which uses it, is green.

## Edge cases
`build_data_dict` implements logic for a wealth of weird edge cases, which never come up in any unit tests and made no sense to me since, as far as I understand, they could only possibly be generated if someone manually crafted a virtual dataset and then expected versioned_hdf5 to read it. Detailed discussion below.

### DIFF
--- a/versioned_hdf5/slicetools.pyx
+++ b/versioned_hdf5/slicetools.pyx
@@ -4,6 +4,7 @@ from functools import lru_cache
 
 import cython
 import h5py
+from h5py import h5d
 
 cimport numpy as np
 
@@ -20,7 +21,7 @@ from libc.stdio cimport FILE, fclose
 from libcpp.vector cimport vector
 
 from versioned_hdf5.cytools import np_hsize_t
-from versioned_hdf5.cytools cimport count2stop, hsize_t, stop2count
+from versioned_hdf5.cytools cimport ceil_a_over_b, count2stop, hsize_t, stop2count
 from versioned_hdf5.tools import asarray
 
 
@@ -103,6 +104,9 @@ cdef extern from "hdf5.h":
         const hsize_t* block,
     ) nogil
 
+# https://github.com/numpy/numpy/blob/b0a52b76c2bebe5c237722589d13bf02affa9c43/numpy/core/include/numpy/ndarraytypes.h#L30
+cdef enum:
+    NPY_MAXDIMS = 32
 
 np.import_array()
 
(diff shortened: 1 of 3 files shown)


versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+    accesses a versioned virtual dataset.
+
+    :param dcpl: the dataset creation property list of the versioned dataset
+    :return: tuple of (slab_indices, slab_offsets) arrays, each with the same shape as
+    the chunks in the virtual dataset.
+    """
+    cdef hsize_t[NPY_MAXDIMS] chunk_size_arr
+    cdef hsize_t[NPY_MAXDIMS] nchunks_arr
+    cdef hsize_t[NPY_MAXDIMS] h5_start
+    cdef hsize_t[NPY_MAXDIMS] scratch  # Dumping ground for unused outputs
```
crusaderky, line 259: Zero cost memory allocation in the stack.
This replaces the vectors used by _spaceid_to_slice (lines 159:162), which are allocated and de-allocated tens of thousands of times in the heap.

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.infer_types(True)
+cpdef tuple[np.ndarray, np.ndarray] build_slab_indices_and_offsets(
+    dcpl,
+    shape: tuple[int, ...],
+    chunk_size: tuple[int, ...],
+):
+    """Build the slab_indices and slab_offsets arrays for a StagedChangesArray that
+    accesses a versioned virtual dataset.
```
crusaderky, line 250: These are explained in #386 -> staged_changes.py -> docstring of StagedChangesArray

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+        assert chunk_size_j > 0
+        chunk_size_arr[j] = chunk_size[j]
+        nchunks_j = ceil_a_over_b(shape[j], chunk_size_j)
+        nchunks_arr[j] = nchunks_j
+        nchunks.append(nchunks_j)
+
+    # Default for missing chunks is to point to slabs[0],
+    # which is a single chunk covered in fill_value
+    cdef np.ndarray slab_indices = np.zeros(nchunks, dtype=np_hsize_t)
+    cdef np.ndarray slab_offsets = np.zeros(nchunks, dtype=np_hsize_t)
```
crusaderky, line 274: Read h#386 -> staged_changes.py -> docstring of StagedChangesArray

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+    indices_strides = slab_indices.strides
+    slab_indices_data = slab_indices.data
+    slab_offsets_data = slab_offsets.data
+
+    with phil:
+        # Newly created sparse datasets completely full of fill_value are not virtual.
+        # FIXME Error getting virtual sources on empty virtual dataset
+        # https://github.com/h5py/h5py/issues/1660
+        if dcpl.get_layout() != h5d.VIRTUAL or 0 in shape:
+            return slab_indices, slab_offsets
```
crusaderky, line 284: Refactored from https://github.com/deshaw/versioned-hdf5/blob/0affe5c0730bc0c2e89b9879e8e0caed8864c4db/versioned_hdf5/wrappers.py#L1403-L1415

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+            for chunk_i in range(virtual_count):
+                # Calculate offset of the chunk, IN BYTES, relative to the start of the
+                # flattened buffer of slab_indices and slab_offsets
+                vspace_id = H5Pget_virtual_vspace(dcpl_id, chunk_i)
+                if vspace_id == H5I_INVALID_HID:
+                    raise HDF5Error()
+                try:
+                    vspace_type = H5Sget_select_type(vspace_id)
+                    if vspace_type != H5S_sel_type.H5S_SEL_HYPERSLABS:
+                        raise ValueError(f"Unexpected {vspace_type=}")
```
crusaderky, line 300: `_spaceid_to_slice` handles this edge case. I cannot think how to possibly generate it short of manually crafting a virtual dataset that is not made by versioned_hdf5. @asmeurer any ideas?

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+
+                    # Get chunk index in the virtual dataset
+                    if H5Sget_regular_hyperslab(
+                        vspace_id,
+                        h5_start,
+                        scratch,  # IGNORE - stride (always 1)
+                        scratch,  # IGNORE - count
+                        scratch,  # IGNORE - block (typically, but not always, 1)
```
crusaderky, line 308: Inconsequential as long as the virtual dataset is contiguous, which is always the case

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+
+                # Calculate offset of the chunk, IN POINTS,
+                # relative to the start of raw_data along axis 0
+                srcspace_id = H5Pget_virtual_srcspace(dcpl_id, chunk_i)
+                if srcspace_id == H5I_INVALID_HID:
+                    raise HDF5Error()
+                try:
+                    srcspace_type = H5Sget_select_type(srcspace_id)
+                    if srcspace_type != H5S_sel_type.H5S_SEL_HYPERSLABS:
+                        raise ValueError(f"Unexpected {srcspace_type=}")
```
crusaderky, line 332: Same as above @asmeurer

versioned_hdf5/slicetools.pyx
```diff
@@ -234,6 +238,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+
+                    # Get coordinates in raw_data
+                    if H5Sget_regular_hyperslab(
+                        srcspace_id,
+                        h5_start,  # Axis 0 = slab_offset; other axes = always 0
+                        scratch,   # IGNORE - stride (always 1)
+                        scratch,   # IGNORE - count
+                        scratch,   # IGNORE - block (typically, but not always, 1)
```
crusaderky, line 340: Inconsequential as long as raw_data is contiguous, which is always the case. Chunk size can be trivially calculated from the global chunk size plus a cut to shape to be applied to the edge chunks

versioned_hdf5/slicetools.pyx
asmeurer: Sorry, I need more context to understand the question you are asking here (it's been a while since I've worked on this codebase).

crusaderky: 4.5 years ago  ~~in a galaxy far away~~, when you originally implemented this function in https://github.com/deshaw/versioned-hdf5/commit/b577ad15c1860f0d3ca00a5fa5581769357a33b7, you covered the edge case of either virtual or source dataspaces being H5S_SEL_ALL instead of H5S_SEL_HYPERSLABS. Shortly afterwards, in https://github.com/deshaw/versioned-hdf5/commit/f3a4e3bd0cd26063c7af36dcb7c31ec37697b20c you added support for H5S_SEL_NONE.

These two use cases never occur in the test suite, and I can't think of any way to reproduce them, short of manually crafting a virtual dataset and then expecting versioned_hdf5 to treat it like a versioned dataset - which IMHO should not be something we care about?

asmeurer: My guess is I was just being defensive. I do know I tried using hyperslabs at some point to improve performance, but it turned out that they were slower in the underlying HDF5 library than what we are doing now.

versioned_hdf5/slicetools.pyx
```diff
@@ -235,6 +239,125 @@ cpdef build_data_dict(dcpl, raw_data_name: str):
...
+                    if H5Sclose(vspace_id) < 0:
+                        raise HDF5Error()
+
+                indices_offset: ssize_t = 0
+                for j in range(ndim):
+                    chunk_idx = h5_start[j] // chunk_size_arr[j]
+                    if h5_start[j] % chunk_size_arr[j] > 0:
+                        raise ValueError("Misaligned chunk")
+                    if chunk_idx >= nchunks_arr[j]:
+                        raise ValueError("chunk_idx too high")
```
peytondmurray, line 322: Are there (possibly contrived) situations where these errors could be raised?

versioned_hdf5/tests/test_slicetools.py
```diff
@@ -47,6 +52,96 @@ def test_spaceid_to_slice(h5file):
...
+    expect[1, 3] = 456
+    expect[2, 0] = 789
+    np.testing.assert_array_equal(virt_dset[:], expect, strict=True)
+    expect_raw = np.array(
+        [
+            [123, 123, 123],
+            [456, 123, 123],
+            [789, 123, 123],
+            [123, 123, 123],
+            [123, 123, 123],  # FIXME Spurious extra chunk
```
peytondmurray, line 132: Is it weird that the spurious chunk in `r1` is at the end of the array?
Seems like there should be a chunk 0 (but not a chunk 3, since it duplicates chunk 0)

`r0` raw data:

```
chunk 0          chunk 1
[123, 123, 123], [123, 123, 123],
[123, 123, 123], [123, 123, 123],

chunk 2          chunk 3
[123, 123, 123], [123, 123, 123],
[123, 123, 123], [123, 123, 123],
```

`r1` raw data:
```

chunk 0          chunk 1
[123, 123, 123], [123, 123, 123],
[123, 123, 123], [456, 123, 123],

chunk 2          chunk 3
[789, 123, 123], [123, 123, 123],
[123, 123, 123], [123, 123, 123],
```

peytondmurray (approved):
Thanks for this :+1:

versioned_hdf5/slicetools.pyx
crusaderky: Not on a round-trip on the current version of versioned_hdf5.

They can come up if someone manually crafts a virtual dataset that kinda looks like a versioned_hdf5 one but it's not quite right. More realistically, if someone in the future reimplements versioned_hdf5 in a different language.

They may also come up potentially if an older, buggy version of versioned_hdf5 wrote malformed files, or if the data model changed and we didn't implement a transition layer.

versioned_hdf5/tests/test_slicetools.py
crusaderky: Yes, it is very weird.

When you commit r0, there is a spurious chunk full of 123 at the beginning of the raw_data dataset.
It is spurious, because nothing in the data_dict points to it. `build_data_dict` raises `ValueError: Not a virtual storage layout (not a virtual storage layout)`. This is preempted by the special case logic from wrappers.py I linked above.

After writing r1, this spurious chunk moves to the *end* of raw_data.

Your diagram above isn't quite right. The actual one is

from the virtual dataset's POV:
```
r0
---
chunk 0 [0:2, 0:3]                 chunk 1 [0:2, 3:5]
not a virtual dataset              not a virtual dataset

chunk 2[2:3, 0:3]                  chunk 3 [2:3, 3:5]
not a virtual dataset              not a virtual dataset

r1
---
chunk 0 [0:2, 0:3]                 chunk 1 [0:2, 3:5]
not in get_virtual_count()         raw_data[0:2, :]
 
chunk 2[2:3, 0:3]                  chunk 3 [2:3, 3:5]         
raw_data[2:4, :]                   not in get_virtual_count()
```
from raw_data's POV:
```
r0
----
chunk 0 [0:2, :]
unused

r1
---
chunk 0 [0:2, :]
virtual_dataset[0:2, 3:5]

chunk 1 [2:4, :]
virtual_dataset[2:3, 0:3] 

chunk 2 [4:6, :]
unused
```

PR #373, opened 2024-09-11
[InMemoryDataset redesign] Add features to subchunk_map

crusaderky:
- Part of #386 
- Blocked by and incorporates #388 

This PR adds extra features, for now unused by the main code, to `subchunk_map.py`:

- New attributes and methods of `IndexChunkMapper`:
  - `n_chunks`
  - `last_chunk_size`
  - `chunks_indexer()`
  - `whole_chunks_idxidx()`
  - `value_view_idx`

PR #386, opened 2024-10-17
Redesign InMemoryDataset data model (libhdf5 direct access)

crusaderky:
This PR has been broken down into several stages - read below.

`versioned_hdf5.wrappers.InMemoryDataset` is plagued by slowness caused by the control code, both because the control code itself is very slow and because it calls `h5py.Dataset.__getitem__` on the raw_data independently for every single chunk.

This PR completely replaces the internal implementation of InMemoryDataset by instead performing a fast inner loop of calls to H5DRead, directly to the libhdf5 C library. The control logic - the generation of the index information needed to perform the transfers, which was previously performed by `build_data_dict` and `as_subchunk_map` - has also been completely rewritten and is now much more aggressively cythonized.

This new design introduces a helper class, `StagedChangesArray`, to which `InMemoryDataset` is now a fairly thin wrapper, and which holds the modified chunks. This class has minimal external API and is completely abstracted from h5py. Internally, all the control logic is encapsulated by a series of `*Plan` classes (`GetItemPlan`, `SetItemPlan`, etc.), each formulating what actions need to be performed in order to evade a user request (`__getitem__`, `__setitem__`, etc.). On each user request,
1. `StagedChangesArray` formulates a plan;
2. executes it, possibly mutating its internal state;
3. discards the plan.

## Other major design changes
In master `__setitem__` first loads all impacted chunks into memory, in the off chance that they are not wholly covered by the parameter index, and then updates them. In this PR, this happens exclusively for the chunks that are not wholly covered by the index of `__setitem__`.
In other words, `a[:] = 1` in master loads the whole dataset into memory if it's not in memory already, whereas in this PR it never touches the disk.

Analogously, in master `resize()` reads the whole dataset from disk. This no longer happens in this PR, which only loads the edge chunks if the shape is not exactly divisible by chunk_size.

There is no longer any direct `__getitem__` call to the virtual dataset. It was found to be exceptionally slow, due to deficiencies in the algorithm in libhdf5 C itself, and has been ditched.
A previous redesign, #370, had to be scrapped due to this limitation.
It remains possible to obtain a plain-h5py virtual dataset by accessing the address
 `_version_data/versions/<version name>/<dataset name>`. Notably, this can be performed in any language that offers libhdf5 bindings, not just python.

There is no longer cache-on-read: any an all caching is performed by libhdf5; the only data in memory within the StagedChangesArray are the chunks that the user intentfully modified.

## Status
- [x] implementation
- [x] all pre-existing unit tests pass
- [x] additional unit tests for the various subsystems (slicetools.pyx, subchunk_map.py)
- [x] additional unit tests for staged_changes.py
- [x] high level documentation
- [x] break PR up into stages to simplify code review
- [x] clean up legacy code
- [x] ad-hoc performance benchmarks
- [ ] review published benchmarks
- [x] demo notebooks

## Demo and benchmakrs
See jupyter notebooks (not to be merged into master) at https://github.com/crusaderky/versioned-hdf5/tree/staged_changes_extras

## PRs chain
This change is in excess of 6000 lines.
For the sake of sanity, it has been broken down into multiple PRs:

- #374
- #378
- #382 
- #385 
- #384 
- #387
- #372
- #391 
- #392 
- #388
- #373
- #390
- #393 
- #394
- #386 (this PR): Integration: hot-swap the implementation of InMemoryDataset. Until this PR, all of the above work lies dormant.
- (blocked by this PR) #395

## Known issues
Slices with step>1, as well as fancy indices which can be locally represented as such, can be slower than in master. The reason is that they're very slow in libhdf5, whereas master was (unconsciously?) side-stepping the issue by loading up whole chunks and then applying the strided slice to the numpy chunks in memory.

This is, technically, a regression only when calling `__getitem__` after `__setitem__`. In master, if you call `__getitem__` on an unmodified dataset, you access the virtual dataset, which is even slower as the issue with step>1 interplays with libhdf5 cache size, and the whole virtual dataset is always larger than the hdf5 cache.

This issue is mitigated, but not solved, by making sure that the [chunk cache](https://docs.h5py.org/en/stable/high/file.html#chunk-cache) is larger than a single chunk. 

Compared to a single contiguous read (potentially followed by strided copy in pure numpy), the performance of reading with step>1 directly from hdf5 is:

- up to 12x slower, if a chunk fits in the cache (1MB by default)
- up to 150x slower, for larger chunks

Master shows a 120x slowdown, regardless of chunk size, but only when reading from the virtual dataset (no changes).

Implementing cache-on-read would effectively work around this issue, but it's not trivial given the current design (the easiest way to do it is to first move hashing into the StagedChangesArray, which makes a lot of sense in its own right anyway).

### DIFF
--- a/versioned_hdf5/versions.py
+++ b/versioned_hdf5/versions.py
@@ -130,7 +130,7 @@ def commit_version(
         shape = None
         if isinstance(data, InMemoryDataset):
             shape = data.shape
-            if data.id._data_dict is None:
+            if not data.staged_changes.has_changes:
                 # The virtual dataset was not changed from the previous
                 # version. Just copy it to the new version directly.
                 assert data.name.startswith(prev_version.name + "/")
@@ -142,7 +142,7 @@ def commit_version(
                 for k, v in data.attrs.items():
                     data_copy.attrs[k] = v
                 continue
-            data = data.id._data_dict
+            data = data.data_dict
         if isinstance(data, dict):
             if chunks[name] is not None:
                 raise NotImplementedError("Specifying chunk size with dict data")
--- a/docs/design.md
+++ b/docs/design.md
@@ -274,6 +274,11 @@ The ndindex library was initially created for versioned-hdf5, in order to make
 index manipulation possible as well as allowing code that passes indices around
 to become much cleaner.
 
+`InMemoryDataset` is implemented as a fairly thin wrapper around
+`StagedChangesArray`, which is abstracted from h5py and holds all the modified
+chunks in memory, presenting them as a numpy-like array, until they are ready to
+be flushed to disk. For details, read [Staging changes in memory](staged_changes).
+
 These wrapper objects all try to emulate the h5py API as closely as possible,
 so that the user can use them just as they would the real h5py objects. Any
 discrepancy between h5py and versioned-hdf5 semantics should be considered a
--- a/versioned_hdf5/replay.py
+++ b/versioned_hdf5/replay.py
@@ -5,7 +5,6 @@ import logging
 import posixpath
 import sys
 from collections.abc import Iterable
-from copy import deepcopy
 from typing import Any
 
 import numpy as np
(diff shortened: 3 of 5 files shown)


crusaderky:
@ArvidJB @peytondmurray This is ready to be played with.

The state is the same as it was in #370 when we abandoned it - the new PR passes all pre-existing tests, but more thorough tests still need to be added, so do expect bugs particularly on edge cases.

versioned_hdf5/versions.py
```diff
@@ -142,7 +142,7 @@ def commit_version(
                 for k, v in data.attrs.items():
                     data_copy.attrs[k] = v
                 continue
-            data = data.id._data_dict
+            data = data.data_dict
```
peytondmurray, line 145: :+1:

peytondmurray (approved):

PR #387, opened 2024-10-20
[InMemoryDataset redesign] ix_with_slices, format_ndindex

crusaderky:
Two new general purpose functions

### DIFF
--- a/versioned_hdf5/tools.py
+++ b/versioned_hdf5/tools.py
@@ -1,3 +1,8 @@
+from __future__ import annotations
+
+from typing import Any
+
+import ndindex
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


versioned_hdf5/tools.py
```diff
@@ -0,0 +1,130 @@
...
+        and hasattr(a, "view")
+    ):
+        # Note that this does not reduce the amount of safety checks:
+        # np.array(-1).astype("u1") doesn't raise and returns 255!
+        return a.view(dtype)
+
+    return a.astype(dtype)
+
+
+def ix_with_slices(*idx: Any, shape: tuple[int, ...]) -> tuple:
```
peytondmurray, line 39: Oh, this is really cool. With this you can more elegantly select hyperrectangles of n-dim arrays. Nice!

peytondmurray (approved):
No complaints here - I look forward to seeing how `ix_with_slices` comes into play in the next PRs :)

Just a few merge conflicts to resolve, then we should merge.

asmeurer:
Both of these things could definitely live in ndindex itself, although it's understandable if you just want to put them here first since ndindex may have more stringent requirements.

PR #388, opened 2024-10-20
[InMemoryDataset redesign] Optimize fancy indices

crusaderky:
- Part of #386 

This PR is made of several commits; the first commits are the incorporated PRs listed above. See only final commit for review.

# PR contents

- Simplify fancy indices down to slices whenever possible - first try to do so globally, and if that's not possible try again locally for each chunk. This will produce an immediate benefit even in the legacy `as_subchunk_map` implementation, as it avoids an expensive intermediate deep-copy.
- Speed up monotonic ascending fancy indices from O(n^2) to O(n*logn)


peytondmurray (approved):
:+1:

PR #389, opened 2024-10-23
Debug libhdf5 version in CI

crusaderky:
- Have the CI workflows print out which version of the libhdf5 .so is actually being loaded up: 1.14 from conda, or an ancient (and unsupported by h5py) 1.10 from ubuntu?
- In benchmarks, move h5py and ndindex from pypi to conda
- In benchmarks, run on the current branch instead of master


.github/workflows/publish_docs.yml
```diff
@@ -33,8 +33,6 @@ jobs:
       # Always checkout master branch, even if triggered by a tag.
       # Otherwise sphinx-multiversion does not pull the master branch.
       - uses: actions/checkout@v4
-        with:
-          ref: master
```
crusaderky, line 37: redundant with `{on: {push: {branches: [master]}}}` above

utils/__init__.py
```diff
@@ -10,3 +10,17 @@ def temp_dir_ctx():
...
     yield tmp_dir
     shutil.rmtree(tmp_dir)
+
+
+def debug_libhdf5_so():
+    """Print out which .so file is actually loaded at runtime for libhdf5"""
+    import h5py
+
+    for row in (
+        subprocess.check_output(["lsof", "-p", str(os.getpid())])
```
crusaderky, line 20: Annoyingly, `psutil.Process().open_files()` filters out all .so files

peytondmurray (approved):
Thanks for this, merging!

PR #391, opened 2024-10-28
[InMemoryDataset redesign] Tweak hypothesis strategies for indices

crusaderky:
Part of #386 

Rework hypothesis strategies in `test_subchunk_map` to allow:
  - starting from size 0
  - generating multiple indices per shape

Both of these features will be leveraged in a later PR to test the StagedChangesArray.


peytondmurray (approved):
Tested locally, thank you!

PR #394, opened 2024-10-28
[InMemoryDataset redesign] StagedChangesArray

crusaderky:
Part of #386

New class StagedChangesArray, which is the new engine underlying the InMemoryDataset. This class is not yet used and will remain dormant until the next PR #386 hot-swaps the engine.

### DIFF
--- a/docs/staged_changes.rst
+++ b/docs/staged_changes.rst
@@ -1,0 +1,428 @@
+Staging changes in memory
+=========================
+This section explains the low level design of ``InMemoryDataset``.
+
+When a ``InMemoryDataset`` is modified in any way, all changed chunks are held in memory
+until they are ready to be committed to disk. To do so, ``InMemoryDataset`` wraps around
+a ``StagedChangesArray`` object, which is a numpy.ndarray-like object.
+
+The ``StagedChangesArray`` holds its data in *slabs*, which is a list of array-like
+objects built as the concatenation of chunks along axis 0, typically with shape
+``(n*chunk_size[0], *chunk_size[1:])``, where n is the number of chunks on the slab.
+This is the same layout as the ``raw_data`` dataset on disk.
+
+The list of slabs is made of three parts:
+
+- The *full slab* is a special slab that is always present and contains exactly one
+  read-only chunk, a broadcasted numpy array full of the fill_value.
+  It's always at index 0.
+- The *base slabs* are array-likes that are treated as read-only. At the moment of
+  writing, when the ``InMemoryDataset`` creates the ``StagedChangesArray``, it passes to
+  it only one base slab, that is the ``raw_data`` ``h5py.Dataset`` - which,
+  conveniently, is a numpy-like object. This slab is typically found at index 1, but may
+  be missing for a dataset that's completely empty.
+- The *staged slabs* are writeable numpy arrays that are created automatically by
+  the ``StagedChangesArray`` whenever there is need to modify a chunk that lies on
+  either the full slab or a base slab.
+
+Two numpy arrays of metadata are used to keep track of the chunks:
+
+- ``slab_indices`` is an array of integers, with the same dimensionality as the virtual
+  array represented by the ``StagedChangesArray`` and one point per chunk, which
+  contains the index of the slab in the ``slabs`` list that contains the chunk.
+- ``slab_offsets`` is an array of identical shape to ``slab_indices`` that contains the
+  offset of the chunk within the slab along axis 0.
+
+e.g.::
+
+  chunk_size[0] = 10
+
+  slab_indices      slab_offsets  slabs[1]              slabs[2]
+  (0 = fill_value)
+
+  1 1 0             30 10  0       0:10 (unreferenced)   0:10 (unreferenced)
+  0 2 0              0 20  0      10:20                 10:20
+  0 2 0              0 10  0      20:30 (unreferenced)  20:30
+                                  30:40
+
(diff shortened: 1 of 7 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -34,6 +34,7 @@
 
 from .backend import DEFAULT_CHUNK_SIZE
 from .slicetools import build_data_dict
+from .staged_changes import StagedChangesArray
```
crusaderky, line 37: While unused until the next PR, this line is necessary here because otherwise pytest would be the first to load the cythonized module and things would get weird and broken.

versioned_hdf5/tests/test_staged_changes.py
```diff
@@ -0,0 +1,705 @@
...
+    fill_value = 42
+
+    if base == "full":
+        base = np.full(shape, fill_value, dtype="u4")
+        arr = StagedChangesArray.full(shape, chunks, fill_value, dtype=base.dtype)
+        assert arr.n_base_slabs == 0
+    elif base == "from_array":
+        base = rng.integers(2**32, size=shape, dtype="u4")
+        # copy base to detect bugs where the StagedChangesArray
+        # accidentally writes back to the base slabs
```
peytondmurray, line 75: :+1:

versioned_hdf5/tests/test_staged_changes.py
```diff
@@ -0,0 +1,705 @@
...
+
+    assert arr.dtype == base.dtype
+    assert arr.fill_value.shape == ()
+    assert arr.fill_value == fill_value
+    assert arr.fill_value.dtype == base.dtype
+    assert arr.slabs[0].dtype == base.dtype
+    assert arr.slabs[0].shape == chunks
+    assert arr.itemsize == 4
+    assert arr.size == np.prod(shape)
+    assert arr.nbytes == np.prod(shape) * 4
```
peytondmurray, line 92: `u4*arr.size` :+1:

versioned_hdf5/tests/test_staged_changes.py
```diff
@@ -0,0 +1,705 @@
...
+    # Test will trip if __getitem__ or __setitem__ perform a
+    # full scan of slab_indices and/or slab_offsets anywhere
+    b = benchmark((2_500, 4_000))
+    np.testing.assert_allclose(b, a, rtol=0.2)
+
+    # 5 million chunks, long rulers
+    # Test will trip if __getitem__ or __setitem__ construct or iterate upon a ruler as
+    # long as the number of chunks along one axis, e.g. np.arange(mapper.n_chunks)
+    c = benchmark((1, 10_000_000))
+    np.testing.assert_allclose(c, a, rtol=0.2)
```
peytondmurray, line 426: How reliable do you expect this to be in CI? If this is flaky I'm guessing there's a way we can measure this with `Mock`, though I suppose there could be something on the h5py side of things that we wouldn't be able to control :thinking:

peytondmurray (approved):
As usual, this looks great, and I'm relying on tests for correctness. Thank you!

docs/staged_changes.rst
```diff
@@ -0,0 +1,428 @@
...
+
+The ``*Plan`` classes are agnostic to data types, never access the actual data (slabs,
+``__getitem__`` return value, or ``__setitem__`` value parameter) and exclusively deal
+in shapes, chunks, and indices.
+
+For debugging purposes, these classes can be generated without executing the method that
+consumes them by calling the ``StagedChangesArray._*_plan()`` methods; this allows
+pretty-printing the list of their instructions e.g. in a Jupyter notebook.
+
+For example, in order to debug what will happen when you call ``dset[2:5, 3:6] = 42``,
```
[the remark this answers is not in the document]
crusaderky: I didn't want to store extra data in the Plan class that would not be used for anything but debugging

versioned_hdf5/tests/test_staged_changes.py
crusaderky: Very reliable.
Even if the hypervisor stole away the CPU while running either use case, which would arbitrarily prolong the wall time, thread time would not increase.

versioned_hdf5/tests/test_staged_changes.py
```diff
@@ -0,0 +1,705 @@
...
+    enlarging the edge chunks without changing the number of chunks.
+    """
+
+    def benchmark(shape):
+        arr = StagedChangesArray.full(shape, (1, 2))
+        # Don't measure page faults on first access to slab_indices and slab_offsets.
+        # In the 10 chunks use case, it's almost certainly reused memory.
+        _ = arr[0, -3:]
+
+        t0 = time.thread_time()
```
crusaderky, line 401:
```suggestion
        # Don't use wall time. The hypervisor on CI hosts can occasionally
        # steal the CPU for multiple seconds away from the VM.
        t0 = time.thread_time()
```

versioned_hdf5/tests/test_staged_changes.py
crusaderky, line 92:
```suggestion
    assert arr.nbytes == np.prod(shape) * 4  # dtype=uint32
```

versioned_hdf5/tests/test_staged_changes.py
crusaderky: None of test_staged_changes.py or staged_changes.py uses h5py. It's all plain numpy.

PR #399, opened 2024-10-31
Move benchmarks to outside repo; update documentation

### LINKED ISSUE #343: Get airspeed velocity benchmarks working again (PyInf#12733)
Not sure when the benchmarks stopped working...
We used to automatically populate https://deshaw.github.io/versioned-hdf5/benchmarks on every commit by running [airspeed velocity](https://github.com/airspeed-velocity/asv).

Let's get this working again!

Also, we may want to update the benchmarks
https://deshaw.github.io/versioned-hdf5/master/performance.html
All of this is probably much faster now?

peytondmurray:
This PR finishes moving the benchmark results to https://github.com/quansight/versioned-hdf5-benchmarks. This will avoid us needing to fight with branch protection rules to store results in this main repository, and anyway it's safer to automate in a separate repository.

I updated the docs to point to the site that hosts the benchmarks, removed the existing results (because they're stored in the other repo), and also bumped the versions of the actions that are needed to publish the docs.

Closes #343.


peytondmurray:
Merging this as low risk changes to the docs.

PR #400, opened 2024-11-19
Remove legacy hacks in ASV benchmarks

crusaderky:
These backwards compatibility hacks are for extremely old versions of versioned_hdf5 which should no longer interest anybody.
The current benchmark suite is being run in CI on >=1.7.0.

### DIFF
--- a/benchmarks/delete_versions.py
+++ b/benchmarks/delete_versions.py
@@ -5,39 +5,9 @@ import tempfile
 import h5py
 import numpy
 
-from versioned_hdf5 import VersionedHDF5File
+from versioned_hdf5 import VersionedHDF5File, delete_versions
 
 filename = "delete_versions_bench.h5"
-
-
-try:
-    from versioned_hdf5 import delete_versions
-except ImportError:
-    from versioned_hdf5.replay import recreate_dataset, swap, tmp_group
-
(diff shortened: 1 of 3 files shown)


benchmarks/inmemoryarraydataset.py
```diff
@@ -1,20 +1,11 @@
 import os
 
 import h5py
+import numpy as np
 
 from versioned_hdf5 import VersionedHDF5File
 from versioned_hdf5.wrappers import InMemoryArrayDataset
 
-try:
-    from versioned_hdf5.wrappers import DatasetWrapper
```
crusaderky, line 9: Introduced in #168 (March 2021)

benchmarks/delete_versions.py
```diff
@@ -5,41 +5,11 @@
...
 import numpy
 
-from versioned_hdf5 import VersionedHDF5File
+from versioned_hdf5 import VersionedHDF5File, delete_versions
 
 filename = "delete_versions_bench.h5"
 
 
-try:
-    from versioned_hdf5 import delete_versions
```
crusaderky, line 14: Introduced in #214 (Dec 2021)

peytondmurray (approved):
Thanks!

PR #403, opened 2024-11-21
Symlinks aren't included in sdist

crusaderky:


.gitignore
```diff
@@ -125,3 +126,6 @@ versioned_hdf5/*.c
 versioned_hdf5/*.cpp
 versioned_hdf5/*.html
 versioned_hdf5/*.so
+# These are temporary copies created by meson.build
+versioned_hdf5/cytools.pyx
+versioned_hdf5/subchunk_map.pyx
```
crusaderky, line 131: Frustratingly, I could not find a way to tell meson to clean up after itself.

peytondmurray (approved):
Thanks for this!

PR #405, opened 2024-12-04
Add support for ellipsis in indices

crusaderky:
- Add support for ... in both `__getitem__` and `__setitem__` indices
- Add test coverage for use case of an ndindex made of one or more scalar indices plus a fancy index, e.g. `[0, [0]]`


peytondmurray (approved):
Thanks for this, not too many lines and a nice feature added :)

PR #418, opened 2025-02-25
Build macos and linux wheels with cibuildwheel

peytondmurray:
This PR adds a workflow for building wheels with `cibuildwheel`. Much of this was adapted from h5py's workflow. I also added an `environment.yml`, which is used for the sdist, and an `environment-windows.yml` with the eventual goal of supporting `msmpi` in the future.

The meson configuration was modified to search for `mpi` and then `ompi` (if the `mpi` dependency was not found). As a last resort, if `mpicc` is on `$PATH` we declare the dependency manually; this was needed to get the manylinux2014 images to be able to build, since `rpm` installs `openmpi-devel` in an awkward location with no `ompi.pc` for `pkg-config` to find.


.github/workflows/pypi_publish.yml
```diff
@@ -12,26 +13,33 @@ on:
...
+        uses: conda-incubator/setup-miniconda@v3
         with:
-          python-version: '3.11'
+          miniconda-version: "latest"
+          activate-environment: build-vhdf
+          environment-file: environment.yml
+          conda-remove-defaults: true
 
-      - name: Install native dependencies
+      - name: Manually sest env variables to allow conda deps to be discoverable
```
crusaderky, line 30:
```suggestion
      - name: Manually set env variables to allow conda deps to be discoverable
```

.github/workflows/pypi_publish.yml
```diff
@@ -12,26 +13,33 @@ on:
...
         run: |
-          set -xe
-          sudo apt update -y
-          sudo apt install -y hdf5-tools libhdf5-dev mpi-default-dev
+          CONDA_PREFIX=$(conda info --base)/envs/build-vhdf
+          echo "CONDA_PREFIX=$CONDA_PREFIX" >> $GITHUB_ENV
+          echo "$CONDA_PREFIX/bin" >> $GITHUB_PATH
+          echo "CPATH=$CONDA_PREFIX/include:$CPATH" >> $GITHUB_ENV
+          echo "LIBRARY_PATH=$CONDA_PREFIX/lib:$LIBRARY_PATH" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$CONDA_PREFIX/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
```
crusaderky, line 38: This is very surprising. `setup-miniconda` _should_ take care of all this.

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +56,104 @@ jobs:
...
 
+  build-wheels:
+    name: 'Wheels: ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.python-version }}'
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: ['3.12']
```
crusaderky, line 67: There should be a different wheel per python version, per OS, per arch
```suggestion
        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
```

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +56,104 @@ jobs:
...
+          - os: macos-latest
+            arch: arm64
+
+    steps:
+    - name: Check out the repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0  # get the non-merge commit for PRs
+
+    # Install openmpi on macos
```
crusaderky, line 82:
```suggestion
```
redundant with title

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +56,104 @@ jobs:
...
+      if: matrix.os == 'macos-latest'
+      run: |
+        brew install open-mpi
+        echo "PKG_CONFIG_PATH=$(brew --prefix)/lib/pkgconfig:$(brew --prefix)/opt/open-mpi/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
+        echo "CPATH=$(brew --prefix)/include:$CPATH" >> $GITHUB_ENV
+        echo "LIBRARY_PATH=$(brew --prefix)/lib:$LIBRARY_PATH" >> $GITHUB_ENV
+
+    # Set up python on ubuntu (build runs in containers, so can't use conda)
+    # Can't use conda on macos, the binaries it installs are invalid according
+    # to meson: "executables created by cpp compiler are not runnable"
```
crusaderky, line 93: Did you try installing the conda-provided compiler tools? (`clangxx_osx-64` on mac and `gxx_linux-64` on linux) 
https://docs.conda.io/projects/conda-build/en/stable/resources/compiler-tools.html

crusaderky, line 93:
```suggestion
    # Don't use conda to build pip wheels
    # (it's going to be messy when you import the packages outside of conda)
```

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +56,104 @@ jobs:
...
+
+    # Use environment-windows.yml because windows only has msmpi, not openmpi
+    - name: 🐍 Set up Python
+      if: matrix.os == 'windows-latest'
+      uses: conda-incubator/setup-miniconda@v3
+      with:
+        miniconda-version: "latest"
+        activate-environment: build-vhdf
+        environment-file: environment-windows.yml
+        conda-remove-defaults: true
```
crusaderky, line 111: What happens if you just use `setup-python` in Windows?
Also the whole windows path is currently commented out at the top of the file

ci/cibw_before_all_macos.sh
```diff
@@ -0,0 +1,30 @@
...
+export HDF5_VERSION="1.14.6"
+export HDF5_DIR="$PROJECT_PATH/cache/hdf5/$HDF5_VERSION-$ARCH"
+# When compiling HDF5, we should use the minimum across all Python versions for a given
+# arch, for versions see for example a more updated version of the following:
+# https://github.com/pypa/cibuildwheel/blob/89a5cfe2721c179f4368a2790669e697759b6644/cibuildwheel/macos.py#L296-L310
+if [[ "$ARCH" == "arm64" ]]; then
+    export MACOSX_DEPLOYMENT_TARGET="11.0"
+else
+    export MACOSX_DEPLOYMENT_TARGET="10.9"
+fi
```
crusaderky, line 20: The github action matrix does not produce wheels for mac-intel though; is this dead code?

pyproject.toml
```diff
@@ -132,3 +132,26 @@ plugins = "numpy.typing.mypy_plugin"
 [[tool.mypy.overrides]]
 module = ["*.tests.*"]
 allow_untyped_defs = true
+
+[tool.cibuildwheel]
+build-verbosity = 1
+build-frontend = { name = "pip", args = ["--only-binary", "numpy"] }
+manylinux-x86_64-image = "ghcr.io/h5py/manylinux2014_x86_64-hdf5"
+manylinux-aarch64-image = "ghcr.io/h5py/manylinux2014_aarch64-hdf5"
```
crusaderky, line 140: This is not triggered by CI though?

.github/workflows/pypi_publish.yml
```diff
@@ -12,26 +13,33 @@ on:
 
 jobs:
   build-sdist:
-    name: Build package
+    name: Build sdist
```
crusaderky, line 16: There's a bug in the sdist package version. This is what I find in the CI artifact:
`versioned_hdf5-0.1.dev1+g5317d17.tar.gz`

crusaderky (changes requested):
This PR adds a lot of dead code for Windows.
If it's unreasonable to get Windows running right away due to outstanding bugs, wouldn't it be better to break the Windows-specific changes out of here and to a follow-up PR?

crusaderky:
I think this would greatly benefit from a smoke test workflow that runs downstream of this, which pip-install on a clean environment the wheels you produced and then runs a hello world. Happy to leave it to a follow-up PR though.

.github/workflows/pypi_publish.yml
peytondmurray: I tried a couple different ways of not setting this, or setting only some of these, but this is the only way I found it works. To be honest I'd be surprised if `setup-miniconda` is setting `CPATH`, but I've spent enough time on this debugging environment variables and github context at this point that I think this is fine. The complexity of the build system is such that it's become much harder to pin down exactly what is strictly necessary and what isn't here.

.github/workflows/pypi_publish.yml
peytondmurray: Thanks for the information - as you mention further down importing packages outside of conda is messy, so I abandoned conda entirely for wheel building.

ci/cibw_before_all_macos.sh
peytondmurray: Leftover from h5py's wheel building stack; now removed.

pyproject.toml
peytondmurray: Removed.

versioned_hdf5/slicetools.pyx
```diff
@@ -19,14 +19,22 @@ from libc.stddef cimport ptrdiff_t, size_t
 from libc.stdio cimport FILE, fclose
 
 from versioned_hdf5.cytools import np_hsize_t
+
```
crusaderky, line 22:
```suggestion
```

versioned_hdf5/slicetools.pyx
```diff
@@ -19,14 +19,22 @@ from libc.stddef cimport ptrdiff_t, size_t
...
 from versioned_hdf5.cytools cimport ceil_a_over_b, count2stop, hsize_t, stop2count
+
```
crusaderky, line 24:
```suggestion
```

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
+
+        # Must use paths relative to the source root here: meson rejects absolute paths
+        CIBW_CONFIG_SETTINGS_WINDOWS: >
+          setup-args="--vsenv"
+          setup-args="-Dmsmpi_libs=['${{ steps.envdata.outputs.MSMPI_PATH }}']"
+          setup-args="-Dzlib_libs=['${{ steps.envdata.outputs.ZLIB_PATH }}']"
+
+    - name: Bundle the hdf5 dlls into the wheels on windows
+      if: runner.os == 'Windows'
+      run: python ./ci/bundle_hdf5_whl.py wheelhouse
```
crusaderky, line 159: I expect this is going to break things. These DLLs are already in the h5py wheels:
```bash
$ unzip -l h5py-3.13.0-cp313-cp313-win_amd64.whl | grep -v "\.py"
Archive:  h5py-3.13.0-cp313-cp313-win_amd64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
  3330560  2025-02-14 02:34   h5py/hdf5.dll
   121344  2025-02-14 02:34   h5py/hdf5_hl.dll
    86016  2017-10-04 10:30   h5py/zlib.dll
```

.github/workflows/pypi_publish.yml
```diff
@@ -12,26 +13,36 @@ on:
...
 
       - name: 🐍 Set up Python
-        uses: actions/setup-python@v4
+        uses: conda-incubator/setup-miniconda@v3
         with:
-          python-version: '3.11'
+          miniconda-version: "latest"
+          activate-environment: build-vhdf
+          environment-file: environment.yml
+          conda-remove-defaults: true
```
crusaderky, line 31: What is the purpose of building the source tarballs with conda (which should have no dependencies other than `build` and `twine`) and everything else with pip? It looks like it's just adding extra complexity to me.

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
           path: ./wheelhouse/*.tar.gz
 
+  build-wheels:
+    name: 'Wheels: ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.python-version }}'
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
crusaderky, line 69: I think this line is redundant with `include`?

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        include:
+          - os: ubuntu-latest
+            arch: x86_64
+          - os: windows-latest
+            arch: AMD64
+          - os: macos-latest
+            arch: arm64
```
crusaderky, line 77: This is potentially problematic. `latest` can change architecture over time; namely macos-latest has recently changed from x86_64 to arm64. I think it's safer to pin the exact VMs.

crusaderky: See matching workflow in h5py:
```yaml
          - os: ubuntu-latest
            arch: x86_64
          - os: ubuntu-24.04-arm
            arch: aarch64
          - os: windows-latest
            arch: AMD64
          - os: macos-14
            arch: arm64
          - os: macos-13
            arch: x86_64
```

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
 
+  build-wheels:
+    name: 'Wheels: ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.python-version }}'
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
```
crusaderky, line 70: Looking at the h5py workflow, they don't use separate python versions in it. Instead, they use a simple "*" (the default in `tiage_build.sh` and then `pypa/cibuildwheel` deals with the multiple iterations.

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
+      with:
+        fetch-depth: 0  # get the non-merge commit for PRs
+        fetch-tags: true # include tags to get correct version from setuptools_scm
+
+    # Don't use conda to build pip wheels
+    # (it's going to be messy when you import the packages outside of conda)
+    - name: 🐍 Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
```
crusaderky, line 91: This happens exclusively in Windows in h5py; I understand that `pypa/cibuildwheel` should deal with it internally?

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
+
+    # MacOS environment setup
+    # libz and libhdf5 require MacOS>=14.0
+    - name: Install openmpi; update env variables on macos
```
crusaderky, line 95: I'm surprised that h5py (which does explicitly use mpi if installed) doesn't need this section, whereas versioned-hdf5 (which doesn't have any mpi-specific code) needs it?

.github/workflows/pypi_publish.yml
```diff
@@ -48,8 +59,122 @@ jobs:
...
+        CIBW_CONFIG_SETTINGS_WINDOWS: >
+          setup-args="--vsenv"
+          setup-args="-Dmsmpi_libs=['${{ steps.envdata.outputs.MSMPI_PATH }}']"
+          setup-args="-Dzlib_libs=['${{ steps.envdata.outputs.ZLIB_PATH }}']"
+
+    - name: Bundle the hdf5 dlls into the wheels on windows
+      if: runner.os == 'Windows'
+      run: python ./ci/bundle_hdf5_whl.py wheelhouse
+
+    - name: Run smoke tests
```
crusaderky, line 161: I think, considering the level of complexity until this point, that it would be a very good idea to start from scratch in a clean environment. You need it anyway:
- build wheels: 1 job per OS/arch; produces 1 wheel per python version
- smoke test: 1 job per OS/arch x 1 job per python version x numpy [1.x, 2.x]

Crucially, the smoke tests DO NOT need you to have libhdf5 or openmpi installed, because the h5py wheels contain everything. If a pure h5py smoke test works, then a versioned-hdf5 smoke test on the same box must work as well.

versioned_hdf5/tests/test_api.py
```diff
@@ -26,6 +26,8 @@
     InMemorySparseDataset,
 )
 
+pytestmark = mark.api
```
crusaderky, line 29: Unused

.github/workflows/pypi_publish.yml
peytondmurray: Yep, I've seen this done either way, so whatever is preferable is fine by me.

versioned_hdf5/slicetools.pyx
```diff
@@ -272,7 +278,7 @@ cdef Exception HDF5Error():
         return RuntimeError("HDF5 error")
 
     cdef char buf[20000]
-    cdef FILE* stream = fmemopen(buf, sizeof(buf), "w")
+    cdef FILE* stream = slicetools_fmemopen(buf, sizeof(buf), "w")
```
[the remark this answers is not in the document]
peytondmurray: We have a check just above this line for `sys.platform` that should prevent us from ever reaching this; it raises a generic `RuntimeError`.

.github/workflows/pypi_publish.yml
peytondmurray: Yep, I'm fine if we remove this hard requirement. IMO it should be optional, and would greatly decrease the complexity of the packaging.

.github/workflows/pypi_publish.yml
peytondmurray: Yes, although `ci/cibw_before_all_windows.sh` contains python commands. I guess we can just conditionally install this on windows only.

.github/workflows/pypi_publish.yml
peytondmurray: Aha, this actually seems like it could be problematic. Let me disable this bundling and see what happens.

.github/workflows/pypi_publish.yml
peytondmurray: Sure, I'll make this pure python and get rid of anything related to conda.

.github/workflows/pypi_publish.yml
peytondmurray: I'm not sure why, but even with `fail-fast: false` and `continue-on-error: true`, only a subset of jobs are launched if I omit this. I'm not sure if that's because if some jobs fail, others don't bother starting (although I would expect that not to be the case from the options above...), or whether there is some other problem with specifying matrices this way. Either way, specifying the matrix as above seems to launch all jobs simultaneously, so I'll keep this as is for now.

.github/workflows/pypi_publish.yml
peytondmurray: > latest can change architecture over time; namely macos-latest has recently changed from x86_64 to arm64. I think it's safer to pin the exact VMs.

Good point. I'll specify the ubuntu arch explicitly.

crusaderky:
@peytondmurray I successfully built a windows wheel. Just as expected, it has the same problem of the mac one, where it feels like there are have two completely separate hdf5 libraries.

Looking at the Linux one now (which works).
Below, both h5py and versioned-hdf5 are installed from wheels.
This _cannot_ be healthy:
```
lib/python3.10/site-packages$ ldd h5py/h5o.cpython-310-x86_64-linux-gnu.so 
        libhdf5-8c29085d.so.310.5.1 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/h5py/../h5py.libs/libhdf5-8c29085d.so.310.5.1 (0x00007ca815400000)
        libhdf5_hl-a45b8ce6.so.310.0.6 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/h5py/../h5py.libs/libhdf5_hl-a45b8ce6.so.310.0.6 (0x00007ca815907000)
        libsz-b97e7bd9.so.2.0.1 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/h5py/../h5py.libs/libsz-b97e7bd9.so.2.0.1 (0x00007ca8158e1000)
        libaec-7e9d22b8.so.0.1.3 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/h5py/../h5py.libs/libaec-7e9d22b8.so.0.1.3 (0x00007ca8158b3000)

lib/python3.10/site-packages$ ldd versioned_hdf5/slicetools.cpython-310-x86_64-linux-gnu.so 
        libhdf5-8c29085d.so.310.5.1 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/versioned_hdf5/../versioned_hdf5.libs/libhdf5-8c29085d.so.310.5.1 (0x00007ff16aa00000)
        libsz-b97e7bd9.so.2.0.1 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/versioned_hdf5/../versioned_hdf5.libs/libsz-b97e7bd9.so.2.0.1 (0x00007ff16aea2000)
        libaec-7e9d22b8.so.0.1.3 => /home/crusaderky/miniforge3/envs/test/lib/python3.10/site-packages/versioned_hdf5/../versioned_hdf5.libs/libaec-7e9d22b8.so.0.1.3 (0x00007ff16ae74000)

lib/python3.10/site-packages$ md5sum h5py.libs/* versioned_hdf5.libs/*
f4dca641ba0aac767d4bebe775ad4305  h5py.libs/libaec-7e9d22b8.so.0.1.3
c3fa212aa26d3a59df141e24ceefd476  h5py.libs/libhdf5-8c29085d.so.310.5.1
0d463e4e728338f418dfc911fd67148b  h5py.libs/libhdf5_hl-a45b8ce6.so.310.0.6
53637b649a5b56a97aa658cbcbabed61  h5py.libs/libsz-b97e7bd9.so.2.0.1
f4dca641ba0aac767d4bebe775ad4305  versioned_hdf5.libs/libaec-7e9d22b8.so.0.1.3
c3fa212aa26d3a59df141e24ceefd476  versioned_hdf5.libs/libhdf5-8c29085d.so.310.5.1
53637b649a5b56a97aa658cbcbabed61  versioned_hdf5.libs/libsz-b97e7bd9.so.2.0.1
```
Neither the Windows nor the MacOS wheels contain a `.libs` directory or the h5py DLLs anywhere though. Continuing investigation...

crusaderky:
@peytondmurray Linux and MacOS are now fully green.

Windows is failing the smoke test with the usual error pointing to a double copy of libhdf5.
A smoking gun is provided by the "repair wheel" step, which fails to notice the dynamic linking to hdf5.dll.
I'm investigating that at the moment but I'm going to timebox it as I suspect diminishing returns.

```
extracting versioned_hdf5-2.0.3.dev123+g76b0bed-cp39-cp39-win_amd64.whl to C:\Users\RUNNER~1\AppData\Local\Temp\tmplrrjba_r
  repairing C:\Users\runneradmin\AppData\Local\Temp\cibw-run-9tr87jwt\cp39-win_amd64\built_wheel\versioned_hdf5-2.0.3.dev123+g76b0bed-cp39-cp39-win_amd64.whl
  finding DLL dependencies
  analyzing package-level extension module versioned_hdf5\cytools.cp39-win_amd64.pyd
  analyzing package-level extension module versioned_hdf5\slicetools.cp39-win_amd64.pyd
  analyzing package-level extension module versioned_hdf5\staged_changes.cp39-win_amd64.pyd
  analyzing package-level extension module versioned_hdf5\subchunk_map.cp39-win_amd64.pyd
  External dependencies to copy into the wheel are
  {'zlib.dll'}
  External dependencies not to copy into the wheel are
  {   'api-ms-win-crt-convert-l1-1-0.dll',
      'api-ms-win-crt-environment-l1-1-0.dll',
      'api-ms-win-crt-filesystem-l1-1-0.dll',
      'api-ms-win-crt-heap-l1-1-0.dll',
      'api-ms-win-crt-math-l1-1-0.dll',
      'api-ms-win-crt-runtime-l1-1-0.dll',
      'api-ms-win-crt-stdio-l1-1-0.dll',
      'api-ms-win-crt-string-l1-1-0.dll',
      'api-ms-win-crt-time-l1-1-0.dll',
      'api-ms-win-crt-utility-l1-1-0.dll',
      'kernel32.dll',
      'python39.dll',
      'vcruntime140.dll'}
  copying DLLs into versioned_hdf5.libs
  copying D:\a\versioned-hdf5\versioned-hdf5\zlib-msvc-x64\build\native\bin_release\zlib.dll -> C:\Users\RUNNER~1\AppData\Local\Temp\tmplrrjba_r\versioned_hdf5.libs\zlib.dll
  mangling DLL names
  repairing cytools.cp39-win_amd64.pyd -> cytools.cp39-win_amd64.pyd
  repairing slicetools.cp39-win_amd64.pyd -> slicetools.cp39-win_amd64.pyd
  repairing staged_changes.cp39-win_amd64.pyd -> staged_changes.cp39-win_amd64.pyd
  repairing subchunk_map.cp39-win_amd64.pyd -> subchunk_map.cp39-win_amd64.pyd
  repairing zlib.dll -> zlib-48f48[517](https://github.com/deshaw/versioned-hdf5/actions/runs/14774711168/job/41480781678#step:10:527)4b14c19986d06e7cfe020930.dll
  patching versioned_hdf5\__init__.py
  updating versioned_hdf5-2.0.3.dev123+g76b0bed.dist-info\RECORD
  repackaging wheel
```

crusaderky:
@peytondmurray I give up. I looked at the modules with a window dependencies tool and, while `zlib` is there, libhdf5 is just not appearing for some reason - neither in the `versioned-hdf5` modules nor in the `h5py` ones from pypi. I'm out of ideas and I'd suggest to just move on.

I've yanked out all Windows-specific changes and I will reopen a DNM draft PR later on to preserve them for posterity, in case someone else wants to give them a stab.

This PR is ready for review and merge.

peytondmurray:
Thanks, being able to build wheels on Linux and MacOS is still a nice improvement. Thanks for giving this a second look, I was losing my mind looking at this beforehand as well :handshake:

PR #420, opened 2025-03-10
[BUG] Change empty dataset detection to work with h5py dev version

### LINKED ISSUE #419: [BUG] Dev version of `h5py` causes issues with `delete_versions`
See https://github.com/deshaw/versioned-hdf5/actions/runs/13510887730/job/37750757706?pr=418 for a log.

```
 =================================== FAILURES ===================================
__________________________ test_delete_empty_dataset ___________________________

vfile = <VersionedHDF5File object "/tmp/pytest-of-runner/pytest-0/test_delete_empty_dataset0/file.hdf5" (mode r+)>

    def test_delete_empty_dataset(vfile):
        """Test that deleting an empty dataset executes successfully."""
        with vfile.stage_version("r0") as sv:
            sv.create_dataset(
                "key0",
                data=np.array([]),
                maxshape=(None,),
                chunks=(10000,),
                compression="lzf",
            )
    
        # Raw data should be filled with fillvalue, but actual current
        # version dataset should have size 0.
        assert vfile.f["_version_data/key0/raw_data"][:].size == 10000
        assert vfile[vfile.current_version]["key0"][:].size == 0
    
        # Create a new version, checking again the size
        with vfile.stage_version("r1") as sv:
            sv["key0"].resize((0,))
        assert vfile.f["_version_data/key0/raw_data"][:].size == 10000
        assert vfile[vfile.current_version]["key0"][:].size == 0
    
        # Deleting a prior version should not change the data in the current version
        delete_versions(vfile, ["r0"])
>       assert vfile.f["_version_data/key0/raw_data"][:].size == 10000
E       assert 0 == 10000
E        +  where 0 = array([], dtype=float64).size

versioned_hdf5/tests/test_replay.py:957: AssertionError
```

Looks like calling `delete_versions` on the original dataset causes the `raw_data` dataset to be erased - but this only seems to affect the development version of h5py, i.e. installing from `git+https://github.com/h5py/h5py`.

peytondmurray:
Fixes #419.

This PR fixes an issue where empty datasets were not being detected correctly, which manifested as a bug when deleting empty datasets. Usually if the user deletes a version of empty dataset, the raw data is not modified because empty datasets don't point to any raw data. If the dataset to be deleted was not empty, this triggers a recreation of the hash table, raw data, and virtual datasets; if the dataset to be deleted was not virtual, we'd skip all of that.

To determine whether a dataset was empty, previously we were checking whether the dataset was virtual, because with previous versions of h5py empty datasets were never allowed to be virtual. With the latest version of h5py empty datasets can now be virtual, meaning our detection logic fails. As a result the raw data, hash table, and other virtual datasets were being modified inadvertently - this caused the test failure.

With this change, we now just look at the length of the dataset to know whether it's empty or not. This _should_ be backwards compatible with existing data, so no migration should be necessary.

### DIFF
--- a/versioned_hdf5/replay.py
+++ b/versioned_hdf5/replay.py
@@ -425,7 +425,7 @@ def _is_empty(f: VersionedHDF5File, name: str, version: str) -> bool:
     bool
         True if the dataset is empty, False otherwise
     """
-    return not f["_version_data/versions"][version][name].is_virtual
+    return f["_version_data/versions"][version][name].len() == 0
 
 
 def _exists_in_version(f: VersionedHDF5File, name: str, version: str) -> bool:


.github/workflows/tests.yml
```diff
@@ -60,7 +60,11 @@ jobs:
         run: pip install git+https://github.com/h5py/h5py
 
       - name: Install versioned-hdf5
-        run: pip install .
+        run: |
+          pip install -v cython meson-python setuptools-scm
+
+          # Build without isolation, so that we target the h5py version we just installed
+          pip install -v . --no-build-isolation
```
crusaderky, line 67: I don't get the comment. What does "target" a version mean? How does this differ from using build isolation?

peytondmurray: Leftover from some debugging.

PR #422, opened 2025-04-07
Accept DTypeLike in `dtype=` parameter in `create_dataset`

crusaderky:
Fix crash with a DTypeLike parameter to `create_dataset`, e.g. `dtype="i4"` instead of `dtype=np.int32`.


peytondmurray (approved):
Look good, but why remove

```toml
plugins = "numpy.typing.mypy_plugin"
```

?

PR #423, opened 2025-04-07
Fix InMemoryArrayDataset dtype

crusaderky:
Fix bugs where `InMemoryArrayDataset.__getitem__` and `__array__` would return the wrong dtype.
Speed up `create_dataset` with list data and explicit dtype, by avoiding a double conversion e.g. list -> np.int64 -> np.int16 (or whatever is the desired dtype).

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -958,6 +958,7 @@ class DatasetLike:
     _fillvalue
     parent (the parent group)
     """
+
     name: str
     shape: tuple[int, ...]
     dtype: np.dtype
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -13,7 +13,7 @@ import warnings
 from collections import defaultdict
 from collections.abc import Iterable
 from functools import cached_property
-from typing import Optional, Union
+from typing import Any
 from weakref import WeakValueDictionary
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -963,7 +967,7 @@ def size(self):
     def fillvalue(self):
         if self._fillvalue is not None:
             return np.array([self._fillvalue], dtype=self.dtype)[0]
-        if self.dtype.metadata:
+        if getattr(self.dtype, "metadata", None):
```
peytondmurray, line 970: What's the difference between this and what was there previously? From what [I can tell from the docs](https://numpy.org/doc/stable/reference/generated/numpy.dtype.metadata.html) it seems like this is always defined on the dtype, although it can have the value of `None`.

versioned_hdf5/wrappers.py
```diff
@@ -1069,36 +1073,26 @@ def as_dtype(self, name, dtype, parent, casting="unsafe"):
...
     @property
     def dtype(self):
-        if self._dtype is not None:
-            return self._dtype
-        return self._array.dtype
+        return self._dtype or self._array.dtype
 
     def __getitem__(self, item):
-        return self.array.__getitem__(item)
+        return np.asarray(self._array[item], dtype=self._dtype)
```
peytondmurray, line 1088: :+1:

peytondmurray (approved):
Nice catch, thank you! :rocket:

versioned_hdf5/wrappers.py
crusaderky: At some point I swear I got an AttributeError. Can't reproduce now. Reverting.

crusaderky: Ah, there you go.
On numpy 1.x, `np.int16` doesn't have a `.metadata` attribute, but `np.asarray(0, dtype=np.int16).dtype` does.

PR #425, opened 2025-04-09
wrappers spring cleaning

crusaderky:
### `modify_metadata`
- Fixed bug where setting `fillvalue=0` would retain the previous fillvalue
- Speed up when the `fillvalue` doesn't change
- Added test for changing  `dtype` on sparse datasets
- Fixed bug when changing `dtype` and `fillvalue` at the same time, which could cause the new `fillvalue` to be transitorily cast to the old `dtype` and overflow, underflow, or lose definition

### `wrappers.py`
- Removed an _astonishing_ amount of dead or useless code
- Simplified code flow
- Removed "lazy" logic in `InMemoryArrayDataset` dtype. It was offering zero performance benefits compared to just eagerly converting input arrays from the users, while being ripe with complexity and bugs.
- Fixed bug that would lead to overflow/underflow/loss of definition after calling `__setitem__` on the initial version of a dataset, caused by the aforementioned lazy dtype logic:
```python
>>> ds = v.create_dataset("x", data=np.asarray([1,2], dtype=np.int8), dtype=np.int64)
>>> ds[0] = 300
OverflowError: Python integer 300 out of bounds for int8
```
- Fixed crashes in `create_dataset` when `chunks` was either a bare integer or `True`
- `create_dataset` will now explicitly disallow `chunks=False` (before it would lead to an uncontrolled crash)
- `create_dataset` will now warn for _all_ ignored kwargs, not just for `maxshape`
- The `.chunks` property would occasionally return an unnecessary `ndindex.ChunkType`; it now always returns a tuple of ints like in h5py
- Improved typing annotations

### DIFF
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -12,7 +12,7 @@ import pytest
 from h5py._hl.filters import guess_chunk
 from numpy.testing import assert_equal
 from packaging.version import Version
-from pytest import mark, raises
+from pytest import raises
 
 from ..api import VersionedHDF5File
 from ..backend import DATA_VERSION, DEFAULT_CHUNK_SIZE
@@ -315,6 +315,26 @@ def test_create_dataset(vfile):
     )
 
 
+def test_create_dataset_warns_for_ignored_kwargs(vfile):
+    """create_dataset() in versioned_hdf5 has a lot less kwargs
+    than the same function in h5py. Test that the extra args are
+    accepted, but ignored with a warning.
+    """
+    match = "parameter is currently ignored for versioned datasets"
+    with vfile.stage_version("v1", "") as group:
+        with pytest.warns(UserWarning, match=match):
+            group.create_dataset("a", data=[1, 2], track_order=True)
+        with pytest.warns(UserWarning, match=match):
+            group.create_dataset("b", data=[1, 2], track_order=False)
+        # None is quietly ignored
+        group.create_dataset("c", data=[1, 2], track_order=None)
+
+        # Quietly ignore maxshape if it's a tuple of Nones
+        with pytest.warns(UserWarning, match=match):
+            group.create_dataset("d", data=[1, 2], maxshape=(3,))
+        group.create_dataset("e", data=[1, 2], maxshape=(None,))
+
+
 def test_changes_dataset(vfile):
     # Testcase similar to those on generate_data.py
     test_data = np.ones((2 * DEFAULT_CHUNK_SIZE,))
@@ -597,7 +617,7 @@ def test_resize_unaligned(vfile):
             assert_equal(group[ds_name][:], np.arange((i + 1) * 1000))
 
 
-@mark.slow
+@pytest.mark.slow
 def test_resize_multiple_dimensions(tmp_path, h5file):
     # Test semantics against raw HDF5
 
--- a/versioned_hdf5/tests/test_api.py
+++ b/versioned_hdf5/tests/test_api.py
@@ -2865,7 +2865,7 @@ def test_other_compression_bad_value(tmp_path, library):
     path = tmp_path / "tmp.h5"
     with h5py.File(path, "w") as f:
         vf = VersionedHDF5File(f)
-        with vf.stage_version("r0") as sv, pytest.raises(ValueError, match="invalid"):
+        with pytest.raises(ValueError, match="invalid"), vf.stage_version("r0") as sv:
             sv.create_dataset(
                 "values",
                 data=np.arange(10),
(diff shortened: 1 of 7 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -200,63 +204,92 @@ def create_group(self, name, track_order=None):
...
+                chunks=chunks,
+            )
+
         self.set_chunks(name, chunks)
-        self.set_compression(name, kwds.get("compression"))
-        self.set_compression_opts(name, kwds.get("compression_opts"))
-        self[name] = data
-        if dtype is not None:
-            self[name]._dtype = dtype
-        return self[name]
```
crusaderky, line 259: `data` could be a np.ndarray in some use cases at this point. Took me a while to figure them out.

versioned_hdf5/wrappers.py
```diff
@@ -377,220 +410,6 @@ def _visit(self, prefix, func):
     # TODO: override other relevant methods here
 
 
-def _make_new_dset(
```
crusaderky, line 380: All this function was doing was
1. convert data from ArrayLike to a np.ndarray
2. in some edge cases, eagerly convert its dtype to the specified dtgype
3. ...that's it. It was creating a dcpl and discarding it, with no effect. All validation was done in `create_dataset`.

versioned_hdf5/wrappers.py
```diff
@@ -660,57 +478,24 @@ def compression_opts(self):
     def compression_opts(self, value):
         self.parent.set_compression_opts(self.name, value)
 
-    def as_dtype(self, name, dtype, parent, casting="unsafe"):
```
crusaderky, line 663: Crashing dead code

versioned_hdf5/wrappers.py
```diff
@@ -660,57 +478,24 @@ def compression_opts(self):
...
-            parent=parent,
-            **init_kwargs,
-        )
-
-        new_dataset.staged_changes = self.staged_changes.astype(dtype, casting=casting)
-        parent[name] = new_dataset
-        return new_dataset
-
     @property
-    def fillvalue(self) -> Any:
```
crusaderky, line 688: Didn't add anything to `super().fillvalue`

versioned_hdf5/tests/test_api.py
```diff
@@ -2845,7 +2865,7 @@ def test_other_compression_bad_value(tmp_path, library):
     path = tmp_path / "tmp.h5"
     with h5py.File(path, "w") as f:
         vf = VersionedHDF5File(f)
-        with vf.stage_version("r0") as sv, pytest.raises(ValueError, match="invalid"):
+        with pytest.raises(ValueError, match="invalid"), vf.stage_version("r0") as sv:
```
crusaderky, line 2868: This PR causes validation of compression to happen upon context exit instead of immediately when calling `create_dataset`. I don't think anybody cares about this change?

crusaderky:
> * `create_dataset` will now warn for _all_ ignored kwargs, not just for `maxshape`

@ArvidJB a heads up, this may lead to some noise particularly for code that runs both on unversioned and versioned datasets. IMHO the warnings are desirable, but nonetheless could you give it a spin and give me some feedback?

peytondmurray (approved):
Are there any implications of removing the `as_dtype` method on `InMemoryArrayDataset`? We no longer track the dtype separately from the underlying array for this type, which makes sense - it's inherently a property of the array.

crusaderky:
> Are there any implications of removing the `as_dtype` method on `InMemoryArrayDataset`? 

I don't believe so. It was a 100% untested and undocumented function.

versioned_hdf5/wrappers.py
```diff
@@ -660,57 +478,24 @@ def compression_opts(self):
...
 
     @property
     def size(self) -> int:
-        return int(np.prod(self.shape))
+        return math.prod(self.shape)
 
     @property
     def chunks(self):
-        return ChunkSize(self.id.chunks)
+        return self.id.chunks
```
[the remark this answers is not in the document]
crusaderky: Worth noting it will be different only for `InMemoryDataset`. It was already a tuple[int, ...] for `InMemoryArrayDataset` and `InMemorySparseDataset`.

> I'm not sure what the implications of this change is
For any code that's made to work both on h5py and versioned-hdf5, none. For versioned-hdf5 specific code, I would be quite surprised if anybody accidentally used some ChunkSize-specific feature.

versioned_hdf5/tests/test_api.py
ArvidJB: I think that's fine, as long as the same exception is raised.

ArvidJB:
> > * `create_dataset` will now warn for _all_ ignored kwargs, not just for `maxshape`
> 
> @ArvidJB a heads up, this may lead to some noise particularly for code that runs both on unversioned and versioned datasets. IMHO the warnings are desirable, but nonetheless could you give it a spin and give me some feedback?

Sounds good to me!

PR #426, opened 2025-04-14
StagedChangesArray to accept AsTypeView as base slab

crusaderky:
Tighten the API surface that a "numpy-like" object must expose in order to be accepted as as a StagedChangesArray base slab. This PR allows h5py `AsTypeView` objects to qualify, and is propaedeutic to implementing NpyStrings datasets (https://github.com/h5py/h5py/pull/2557)

### DIFF
--- a/versioned_hdf5/tools.py
+++ b/versioned_hdf5/tools.py
@@ -4,16 +4,19 @@ from typing import Any
 
 import ndindex
 import numpy as np
+from numpy.typing import ArrayLike, DTypeLike
+
+from .typing_ import ArrayProtocol
 
 
-def asarray(a, dtype=None):
-    """Variant of np.asarray(a, dtype=dtype), with two differences:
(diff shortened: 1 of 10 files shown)


versioned_hdf5/tools.py
```diff
@@ -33,7 +33,9 @@ def asarray(a, dtype=None):
         # np.array(-1).astype("u1") doesn't raise and returns 255!
         return a.view(dtype)
 
-    return a.astype(dtype)
+    if hasattr(a, "astype"):
+        return a.astype(dtype)
+    return np.asarray(a, dtype=dtype)
```
crusaderky, line 38: I could have also added
```python
if hasattr(a, "__array_function__`):
    return np.asarray(a, dtype=dtype, like=a)
```
but it wasn't terribly interesting as a use case in real life.

peytondmurray (approved):
Thanks for this!

PR #428, opened 2025-04-29
Move tests to top level

crusaderky:
Use absolute imports everywhere and move `versioned_hdf5/tests` to the root directory.

This PR has two benefits:
- run tests suite without needing to do weird machinations with symlinks for the pip-installed modules. All modules are imported from the repo dir and, when that fails (for pyx files), from the pip-installed location.
- it is now possible to just remove the versioned_hdf5 directory to be extra sure that all tests are running from the pip-installed location.


peytondmurray (approved):
Thanks, looks good :tada:

PR #429, opened 2025-04-30
Overhaul debug_libhdf5_so

crusaderky:
- Add Windows support
- Allow debugging which libhdf5 is actually loaded depending on what you import first, h5py or versioned_hdf5.
  It *should* always be the same but that's just in virtue of the fact that
  - `import h5py` appears before `import versioned_hdf5.slicetools`, and
  - `h5py` eagerly imports its Cython implementation modules.
- Simplify packaging


peytondmurray (approved):
:rocket: Thank you!

PR #433, opened 2025-06-05
Partial lint pass

crusaderky:
A very incomplete linters review.
Still doesn't pass mypy or ruff, but it's a bit closer than before now.


docs/conf.py
```diff
@@ -9,14 +9,12 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#
-import os
-import sys
 
+# import os
+# import sys
```
peytondmurray, line 14: I guess these can be removed if they are unused?

pyproject.toml
```diff
@@ -73,27 +73,28 @@ select = [
 ignore = [
     "D104", # Missing docstring in public package
     "D100", # Missing docstring in public module
-    "D211", # No blank line before class
     "D213", # Multiline summary second line
     "PD901", # Avoid using 'df' for pandas dataframes. Perfectly fine in functions with limited scope
-    "ANN201", # Missing return type annotation for public function (makes no sense for NoneType return types...)
-    "ANN101", # Missing type annotation for `self`
```
peytondmurray, line 80: :+1:

pyproject.toml
```diff
@@ -73,27 +73,28 @@ select = [
...
     "ANN002", # Missing type annotation for `*args`
     "ANN003", # Missing type annotation for `**kwargs`
     "D105", # Missing docstring in magic method
     "D203", # 1 blank line before after class docstring
-    "D204", # 1 blank line required after class docstring
     "D413", # 1 black line after parameters
     "SIM108", # Simplify if/else to one line; not always clearer
-    "D206", # Docstrings should be indented with spaces; unnecessary when running ruff-format
-    "E501", # Line length too long; unnecessary when running ruff-format
-    "W191", # Indentation contains tabs; unnecessary when running ruff-format
```
peytondmurray, line 91: We're still using `ruff format` here, should these be kept? See https://docs.astral.sh/ruff/rules/tab-indentation/#formatter-compatibility

peytondmurray (approved):
Hey, thanks for this - this cleanup is a long time coming :rocket:

pyproject.toml
crusaderky: I just swapped the running order and run ruff-format first, so these lines become inconsequential

PR #435, opened 2025-06-17
Another partial linters pass

crusaderky:
Follow-up to #433. Some more tweaks. Number of ruff errors going very slowly down.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1446,7 +1446,7 @@ def test_group_chunks_compression(vfile):
 
 
 def test_closes(vfile):
-    data = np.ones((DEFAULT_CHUNK_SIZE,))
+    data = np.ones(DEFAULT_CHUNK_SIZE)
 
     with vfile.stage_version("version1") as g:
         g.create_dataset("test_data", data=data)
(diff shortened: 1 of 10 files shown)


tests/test_api.py
```diff
@@ -1453,9 +1461,9 @@ def test_closes(vfile):
 
     assert vfile._closed is True
     assert vfile.closed is True
-    raises(AttributeError, lambda: vfile.f)
-    raises(AttributeError, lambda: vfile._version_data)
-    raises(AttributeError, lambda: vfile._versions)
+    pytest.raises(AttributeError, lambda: vfile.f)
```
peytondmurray, line 1464: Elsewhere `raises` is being converted to `with pytest.raises(...):` context manager, should the same approach be used here?

tests/test_api.py
```diff
@@ -2527,7 +2539,8 @@ def test_rebuild_hashtable_chunk_reuse_unicode(tmp_path, caplog):
 
 
 def test_rebuild_hashtable_chunk_reuse_multi_dim(tmp_path, caplog):
-    """Test that the correct chunks are used after rebuilding the tables for a multi-dimensional array."""
+    """Test that the correct chunks are used after rebuilding the tables for a
+    multi-dimensional array."""
```
peytondmurray, line 2543:
```suggestion
    multi-dimensional array.
    """
```

I think multiline docstrings should have the terminating `"""` on a separate line.

peytondmurray (approved):
Thanks for this - this kind of cleanup is much needed with this old project!

PR #436, opened 2025-06-19
Fix hashing validation with mix of str and bytes

crusaderky:
This is part of NpyStrings work.

Fix bug where the `ENABLE_CHUNK_REUSE_VALIDATION` flag would trigger a false positive when the first element of an object array is bytes, but then later on there are str elements.

### DIFF
--- a/tests/test_hashtable.py
+++ b/tests/test_hashtable.py
@@ -8,6 +8,8 @@ from versioned_hdf5 import VersionedHDF5File
 from versioned_hdf5.backend import create_base_dataset
 from versioned_hdf5.hashtable import Hashtable
 
+from .conftest import assert_slab_offsets
+
 
 def test_hashtable(h5file):
     create_base_dataset(h5file, "test_data", data=np.empty((0,)))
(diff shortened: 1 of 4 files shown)


tests/test_hashtable.py
```diff
@@ -243,63 +245,71 @@ def data_version_3_hash(self, data: np.ndarray):
...
-    with h5py.File(filename, mode="r+") as f:
-        vf = VersionedHDF5File(f)
-        with vf.stage_version("r1") as sv:
-            sv["values"] = np.concatenate((data, data))
+    with vfile.stage_version("r0") as sv:
+        sv.create_dataset("values", data=data, chunks=(6,))
+    with vfile.stage_version("r1") as sv:
+        sv["values"] = np.concatenate((data, data))
+    with vfile.stage_version("r2") as sv:
+        assert_slab_offsets(sv, "values", [0, 0])
```
crusaderky, line 256: Before, there was no check that the chunks were actually reused!

peytondmurray: Good catch!

PR #437, opened 2025-06-20
Tweak test_api::test_scalar_dataset

crusaderky:
Part of NpyStrings work.
Tweak a test for scalar datasets, adding use cases for `h5py.string_dtype()` and `np.generic`.


crusaderky:
> use the `vfile` fixture for brevity

Good point. Done.

PR #438, opened 2025-06-20
StagedChangesArray: CoW; `astype` hot-swaps base slabs; writeable flag

crusaderky:
This is part of NpyStrings work.

Introduce some new features in StagedChangesArray, not yet leveraged by wrappers.py:

- Copy-on-Write (CoW) mechanics:
  - `copy()` does not actually deep-copy staged slabs until the first write access
  - `astype()` does not actually convert staged slabs until the first read or write access
  - `refill()` eagerly deep-copies and refills staged slabs that actually contain the previous `fill_value`, staged slabs that don't contain it aren't copied until the first write access.
- `astype()` now has the option to hot-swap the base slabs; this will allow hot-swapping the raw dataset) with a h5py AsTypeView of it.
- Added a global `.writeable` flag, that prevents all mutating operations much like `ndarray.flags.writeable`.

### DIFF
--- a/versioned_hdf5/staged_changes.py
+++ b/versioned_hdf5/staged_changes.py
@@ -40,7 +40,6 @@ if cython.compiled:  # pragma: nocover
 
 
 T = TypeVar("T", bound=np.generic)
-Casting = Literal["no", "equiv", "safe", "same_kind", "unsafe"]
 
 
 class StagedChangesArray(MutableMapping[Any, T]):
@@ -159,6 +158,9 @@ class StagedChangesArray(MutableMapping[Any, T]):
     #: Staged slabs start at index n_base_slabs + 1.
     n_base_slabs: int
 
+    #: Set to False to disable mutating methods
+    writeable: bool
+
     __slots__ = tuple(__annotations__)
 
     def __init__(
@@ -184,6 +186,7 @@ class StagedChangesArray(MutableMapping[Any, T]):
 
         self.shape = shape
         self._resized = False
+        self.writeable = True
 
         dtype = base_slabs[0].dtype if base_slabs else None
         if fill_value is None:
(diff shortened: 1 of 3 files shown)


versioned_hdf5/staged_changes.py
```diff
@@ -277,13 +280,18 @@ def n_staged_slabs(self) -> int:
         """Number of staged slabs containing modified chunks"""
         return self.n_slabs - self.n_base_slabs - 1
 
+    @property
+    def staged_slabs_start(self) -> int:
```
peytondmurray, line 284: :+1:

versioned_hdf5/staged_changes.py
```diff
@@ -394,12 +427,42 @@ def _load_plan(self, copy: bool = True) -> LoadPlan:
...
+
+        if writeable:
+            # dst_slab is either a staged slab or the return value of __getitem__
+            assert isinstance(slab, np.ndarray)
+            if not slab.flags.writeable:
+                assert idx is not None  # __getitem__ return value is always writeable
+                assert idx > self.n_base_slabs  # Only staged slabs are writeable
+                # There was a previous call to copy()
+                slab = slab.copy()
+                self.slabs[idx] = slab
```
peytondmurray, line 464: Are there instances in which both of these would be hit, resulting in an extra (needless) copy of the slab? For example if both `astype()` and `copy()` were called previously?

peytondmurray (approved):
Small clarification. Otherwise this looks ready!

versioned_hdf5/staged_changes.py
crusaderky: When `asarray` performs a deep copy in the 451:454 branch, the returned array is always writeable, so the second branch 460:464 is not executed.

There is one special case in `asarray` for uint<->int conversion, which is performed as a view. In that case, you may hit both branches: first `asarray` will return a read-only uint view of a read-only int buffer (or vice versa), which is cheap, then `.copy()` will deep-copy the buffer, which is the only expensive bit.

Issue #440, opened 2025-06-24
`resize()` can be slow and cause huge memory usage on InMemoryArrayDataset


crusaderky:
Calling `.resize()` to enlarge a dataset and then commit, without fully filling the new area first, takes potentially a huge amount of RAM and is extremely slow for InMemoryArrayDataset. 

`InMemoryArrayDataset` is used:
- immediately after `create_dataset(data=...)`, before committing first
- after opening any dataset and subsequently calling `__setitem__` to update its whole area.

The reason is quite simple: for all other datasets (`h5py.Dataset`, `InMemoryDataset`, `InMemorySparseDataset`) the new chunks created by `resize()` are empty and are commited as such - which means no hashing is needed.
On `InMemoryArrayDataset`, however, the new chunks are real and full of the fillvalue. Upon committing, they all need to be hashed and deduplicated.

This is a non-issue if the new chunks created by `.resize()` are all filled by `__setitem__` immediately afterwards, before committing the version.

PR #444, opened 2025-06-27
fix(440): implement lazy resize in InMemoryArrayDataset to reduce mem…

VinayakTiwari1103:
…ory usage

# Improved Memory Management in `InMemoryArrayDataset`

This PR addresses **Issue #440**, significantly reducing memory usage and improving performance when calling `resize()` on `InMemoryArrayDataset`.

## Key Changes

- **Lazy Allocation**  
  The underlying array is now only materialized when accessed or modified (via `__getitem__`, `__setitem__`, or `__array__`).

- **Efficient `resize()`**  
  `resize()` updates the logical shape without immediate memory allocation. Memory is only used when data is actually written.

- **Correct `fillvalue` Handling**  
  Behavior of `fillvalue` is preserved during both reads and writes.

- **Supports Growth & Shrink**  
  Handles both increasing and decreasing dataset size safely.

## Benefits

- **Before:**  
  `resize()` would eagerly fill all new regions, leading to high memory usage.

- **After:**  
  Memory is only allocated when data is accessed, making it ideal for sparse datasets and large-scale workloads.

## Notes

- This update is **isolated** to `InMemoryArrayDataset`.  
- All lazy allocation logic is encapsulated in a private `_ensure_allocated()` method to ensure safety and consistency.

Want to see it in action?  
## Test cases are available and can be added in a follow-up PR.

image

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -818,72 +818,82 @@ class DatasetLike:
 class InMemoryArrayDataset(DatasetLike):
     """
     Class that looks like a h5py.Dataset but is backed by an array
+    with support for lazy allocation.
     """
 
     def __init__(
         self,
         name: str,
-        array: np.ndarray,
+        array: np.ndarray | None,
         *,
         parent: InMemoryGroup,
         fillvalue: Any | None = None,
         chunks: tuple[int, ...] | None = None,
+        shape: tuple[int, ...] | None = None,
+        dtype: DTypeLike | None = None,
     ):
         self.name = name
-        self._array = array
         self.attrs = {}
         self.parent = parent
         self._fillvalue = fillvalue
+
+        self._array = array
+        self._lazy_shape = array.shape if array is not None else shape
+        self._lazy_dtype = array.dtype if array is not None else np.dtype(dtype)
+        if self._array is None and (self._lazy_shape is None or self._lazy_dtype is None):
+            raise ValueError("Either array or (shape and dtype) must be provided")
+
         if chunks is None:
             chunks = parent.chunks[name]
         self.chunks = chunks
 
+    def _ensure_allocated(self):
+        if self._array is None:
+            self._array = np.full(self._lazy_shape, self.fillvalue, dtype=self._lazy_dtype)
+
     @property
     def shape(self) -> tuple[int, ...]:
-        return self._array.shape
+        return self._array.shape if self._array is not None else self._lazy_shape
 
     @property
     def dtype(self) -> np.dtype:
-        return self._array.dtype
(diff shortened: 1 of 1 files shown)


VinayakTiwari1103:
@quarl @asmeurer @scopatz @takluyver please review this PR

versioned_hdf5/wrappers.py
```diff
@@ -818,72 +818,82 @@ def compression_opts(self, value):
...
         self.name = name
-        self._array = array
         self.attrs = {}
         self.parent = parent
         self._fillvalue = fillvalue
+
+        self._array = array
+        self._lazy_shape = array.shape if array is not None else shape
+        self._lazy_dtype = array.dtype if array is not None else np.dtype(dtype)
+        if self._array is None and (self._lazy_shape is None or self._lazy_dtype is None):
```
crusaderky, line 843: array will never be None.
An InMemoryArrayDataset is created in two use cases:
1. a user calls `create_dataset` and passes an array as the `data=` parameter;
2. a user calls `stage_version` and then opens a pre-existing dataset, which creates a `InMemoryDataset`; then it calls `__setitem__` on it to update everything. `DatasetWrapper.__setitem__` hot-swaps the `InMemoryDataset` with a new `InMemoryArrayDataset`.

versioned_hdf5/wrappers.py
```diff
@@ -818,72 +818,82 @@ def compression_opts(self, value):
...
-            ]
-            self._array = new_array
+            old_shape = self._array.shape
+            if all(new <= old for new, old in zip(size, old_shape)):
+                self._array = self._array[tuple(slice(0, s) for s in size)]
+            else:
+                new_array = np.full(size, self.fillvalue, dtype=self._array.dtype)
+                slices = tuple(slice(0, min(n, o)) for n, o in zip(size, old_shape))
+                new_array[slices] = self._array[slices]
+                self._array = new_array
```
crusaderky, line 895: This does not do what the PR suggests it would do. It is still eager, except in the non-existing use case where the InMemoryArrayDataset was created with None.

crusaderky (commented):
This PR is not fit for purpose.
First, it does not deliver on what it is designed to do, as it just creates a new use case which is never actually triggered.
Second, even if it did (which would require more work), it could exclusively fix #440 in the narrow use case where a user calls `resize()` and then commits immediately afterwards, without any changes. If a single point is changed, we revert back to the original problem.

I advise to retire this PR without merging it. My idea was to reuse the `DatasetWrapper.__setitem__` pattern to implement `DatasetWrapper.resize`, which would hot-swap a `InMemoryArrayDataset` for a `InMemorySparseDataset`.

Finally, #427 is in advanced review state and is going to impact heavily this area of code. I advise holding off until then.

PR #445, opened 2025-07-02
NpyStrings benchmarks

crusaderky:
- Blocked by and incorporates https://github.com/deshaw/versioned-hdf5/pull/442.
- #427 enables dtype="T" and fixes `astype` test failures

See only final commit.
**tip:** you can pip-install a different branch (#427), switch back to this branch, and run `asv run --python=same --quick`

### DIFF
--- a/benchmarks/strings.py
+++ b/benchmarks/strings.py
@@ -1,0 +1,90 @@
+import h5py
+
+from .common import Benchmark, require_npystrings
+
+
+class TimeStrings(Benchmark):
+    """Benchmark for string dtypes"""
+
+    params = [
+        ["h5py", "versioned_hdf5"],
+        ["S", "O", "T"],
+        [1, 8, 64, 256],
+    ]
+    param_names = ["library", "dtype", "max_nchars"]
+
+    SHAPE = (200, 200)
+    CHUNKS = (100, 100)
+
+    def setup(self, library, dtype, max_nchars):
+        super().setup()
+        if dtype == "S":
+            dtype = f"S{max_nchars}"
+        elif dtype == "O":  # object strings
+            dtype = h5py.string_dtype()
+        elif dtype == "T":  # NpyStrings a.k.a. StringDType
+            require_npystrings()
+
+        self.data1 = self.rand_strings(self.SHAPE, 0, max_nchars, dtype)
+        self.data2 = self.rand_strings(self.SHAPE, 0, max_nchars, dtype)
+
+        if library == "h5py":
+            self.file.create_dataset("x", data=self.data1, chunks=self.CHUNKS)
+            self.reopen()
+            self.ds = self.file["x"]
+        else:
+            with self.vfile.stage_version("v0") as sv:
+                sv.create_dataset("x", data=self.data1, chunks=self.CHUNKS)
+            self.reopen()
+            self.ctx = self.vfile.stage_version("v1")
+            sv = self.ctx.__enter__()
+            self.ds = sv["x"]
+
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Install versioned-hdf5
+        run: pip install .
+
+      - name: Dump conda environment
+        run: mamba env export
+
+      - name: Setup asv
+        run: |
+          # Work around https://github.com/airspeed-velocity/asv/issues/1508
+          sed -i "s/master/$(git rev-parse --abbrev-ref HEAD)/" asv.conf.json
```
peytondmurray, line 64: aiaiai, sorry about this. I also had no luck with this when setting up `versioned-hdf5-benchmarks`. I think you can also do this by specifying `HASHFILE:<some_file_containing_hashes>` for the range arg, but it looks like this works too.

benchmarks/wrappers.py
```diff
@@ -0,0 +1,158 @@
...
+        if kind != "h5py.Dataset":
+            self.ctx.__exit__(None, None, None)  # commit
+            del self.ctx
+
+    def time_resize_smaller(self, kind):
+        self.assert_clean_setup()
+        self.ds.resize((10, 10, 10))
+        if kind != "h5py.Dataset":
+            self.ctx.__exit__(None, None, None)  # commit
+            del self.ctx
```
peytondmurray, line 78: Taken care of by `teardown`? Same comment in `time_setattr` and `time_resize_bigger`.

```suggestion
```

benchmarks/wrappers.py
```diff
@@ -0,0 +1,158 @@
...
+    params = [
+        "h5py.Dataset",
+        "InMemoryArrayDataset",
+        "InMemoryDataset",
+        "InMemorySparseDataset",
+    ]
+    param_names = ["kind"]
+
+    def setup(self, kind):
+        super().setup()
```
peytondmurray, line 16: This writes a fresh file each benchmark 👍

benchmarks/strings.py
```diff
@@ -0,0 +1,90 @@
...
+    CHUNKS = (100, 100)
+
+    def setup(self, library, dtype, max_nchars):
+        super().setup()
+        if dtype == "S":
+            dtype = f"S{max_nchars}"
+        elif dtype == "O":  # object strings
+            dtype = h5py.string_dtype()
+        elif dtype == "T":  # NpyStrings a.k.a. StringDType
+            require_npystrings()
```
peytondmurray, line 26: 👍

benchmarks/wrappers.py
```diff
@@ -0,0 +1,158 @@
...
+            else:
+                raise AssertionError("unreachable")
+
+        else:
+            raise AssertionError("unreachable")
+
+    def time_commit(self, kind, chunks):
+        self.assert_clean_setup()
+        self.ctx.__exit__(None, None, None)
+        self.file.close()
```
peytondmurray, line 158: Should this go into a `teardown` method? Does it even contribute much overhead at all, or is it negligible?

benchmarks/wrappers.py
crusaderky: The point is that I want to include the commit phase in the metrics. Otherwise it makes no sense to compare against pure h5py, which writes to disk within the same command. I'll add a comment about it.

benchmarks/wrappers.py
crusaderky: You're right, it's inconsequential

PR #446, opened 2025-07-07
Efficiently enlarge InMemoryArrayDataset

### LINKED ISSUE #440: `resize()` can be slow and cause huge memory usage on InMemoryArrayDataset

crusaderky:
- Closes #440
- Supersedes #444

Hot-swap InMemoryArrayDataset with a InMemorySparseDataset when the user tries to enlarge it.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1641,12 +1641,12 @@ def test_check_committed(vfile):
     # Incorrectly uses g from the previous version (InMemoryArrayDataset)
     with pytest.raises(ValueError, match="committed"):
         with vfile.stage_version("version2"):
-            assert isinstance(g["test_data"], InMemoryArrayDataset)
+            assert isinstance(g["test_data"].dataset, InMemoryArrayDataset)
             g["test_data"][0] = 1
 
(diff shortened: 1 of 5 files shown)


peytondmurray (approved):
Thanks, I tested this locally and all seems well :)

PR #454, opened 2025-08-12
Clean up blosc compression tests

crusaderky:


.github/workflows/tests.yml
```diff
@@ -69,7 +69,7 @@ jobs:
       - name: Install versioned-hdf5 test packages and extra h5py compression types
         run: |
           pip install .[test]
-          pip install hdf5plugin tables psutil
```
crusaderky, line 72: read below for explanation

tests/test_api.py
```diff
@@ -2968,67 +2966,46 @@ def test_verify_string_chunk_reuse_bytes_one_dimensional(tmp_path):
...
+        )
 
 
-@pytest.mark.parametrize(
-    ("library"),
-    [
-        "hdf5plugin",
-        "tables",
-    ],
-)
```
crusaderky, line 3003: This did not make sense.

From my reading of hdf5plugin, which .so files are injected into libhdf5 when both tables and hdf5plugin is undefined and depends on import order.
In order to be sure to test both modules, one would need to run the test in a subprocess with method=spawn and be extra careful to import only one of the two.
IMHO this would be overkill, as both hdf5plugin and pytables interact in exactly the same way with versioned-hdf5. So I chose hdf5plugin, which has the additional user-friendly class API.

peytondmurray (approved):
Thanks for this 🚀

PR #455, opened 2025-08-12
Fix bugs around compression parameters and properties

crusaderky:
- Fix bug where the `.compression` and `.compression_opts` properties of staged datasets would incorrectly return None
- Fix bug where passing a path to `create_dataset` would silently disregard the `compression` and `compression_opts` parameters.

e.g. this would work as intended:
```python
with vf.stage_version("r1") as v:
    g = v.create_group("a/b")
    ds = g.create_dataset("c", compression="gzip")
```
while this would disregard the compression parameter:
```python
with vf.stage_version("r1") as v:
    ds = v.create_dataset("a/b/c", compression="gzip")
```

### DIFF
--- a/versioned_hdf5/wrappers.py
+++ b/versioned_hdf5/wrappers.py
@@ -144,10 +144,9 @@ class InMemoryGroup(Group):
             wrapped_dataset = self._data[name] = DatasetWrapper(
                 InMemoryDataset(obj.id, parent=self)
             )
-            self.set_compression(name, wrapped_dataset.dataset.id.raw_data.compression)
-            self.set_compression_opts(
-                name, wrapped_dataset.dataset.id.raw_data.compression_opts
-            )
+            raw_data = wrapped_dataset.dataset.id.raw_data
+            self.set_compression(name, raw_data.compression)
+            self.set_compression_opts(name, raw_data.compression_opts)
         elif isinstance(obj, Group):
             self._subgroups[name] = InMemoryGroup(obj.id)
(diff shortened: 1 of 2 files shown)


versioned_hdf5/wrappers.py
```diff
@@ -376,42 +375,32 @@ def set_chunks(self, item, value):
     def compression(self):
         return self._compression
 
-    def set_compression(self, item, value):
-        full_name = item
+    def set_compression(self, name, value):
+        _, basename = posixpath.split(name)
+        full_name = basename
```
peytondmurray, line 380:
```suggestion
        full_name = posixpath.basename(name)
```

works too.

versioned_hdf5/wrappers.py
```diff
@@ -531,8 +520,32 @@ def _astype_impl(self, dtype: np.dtype, writeable: bool) -> MutableArrayProtocol
...
-# Note: BufferMixin methods override those from Dataset
-class InMemoryDataset(BufferMixin, Dataset):
+class CompressionMixin:
+    name: str
+    parent: InMemoryGroup
+
+    @property
+    def compression(self):
+        _, basename = posixpath.split(self.name)
+        return self.parent.compression[basename]
```
peytondmurray, line 530:
```suggestion
        return self.parent.compression[posixpath.basename(self.name)]
```

Any reason not to use posixpath.basename?

peytondmurray (approved):
Looks like a nice deduplication with the `CompressionMixin` 👍

Issue #449, opened 2025-07-18
Support shuffle parameter in `create_dataset`


wuke-deshaw:
versioned-hdf5 2.0.2 silently ignores shuffle=True (and other parameters like fletcher32) specified during `create_dataset`.
The latest code [turns this into a warning](https://github.com/deshaw/versioned-hdf5/blob/dd03443942c8b2f00635e62bee221e2788428151/versioned_hdf5/wrappers.py#L232-L237), but ideally we should support this.

(Internal ticket: https://desflow.nyc.deshaw.com/Q/PyInf/15920)

Reproducer:

```python
import h5py
import numpy as np
from   versioned_hdf5           import VersionedHDF5File
from   pathlib                  import Path

path = Path('./test_versioned.h5')
if path.exists():
    path.unlink()

with h5py.File(path, "w") as base_file:
    vfile = VersionedHDF5File(base_file)
    with vfile.stage_version('v1') as group:
        group.create_dataset(
            "data",
            None,
            maxshape=(None, None),
            chunks=(1000, 1000),
            compression="lzf",
            shuffle=True,
            data=np.random.rand(1000, 1000),
        )
```

```console
$ h5dump -pH --dataset=/_version_data/data/raw_data test_versioned.h5
HDF5 "test_versioned.h5" {
DATASET "/_version_data/data/raw_data" {
   DATATYPE  H5T_IEEE_F64LE
   DATASPACE  SIMPLE { ( 1000, 1000 ) / ( H5S_UNLIMITED, 1000 ) }
   STORAGE_LAYOUT {
      CHUNKED ( 1000, 1000 )
      SIZE 8000000 (1.000:1 COMPRESSION)
   }
   FILTERS {  # <------------ no shuffle filter on the raw_data dataset
      USER_DEFINED_FILTER {
         FILTER_ID 32000
         COMMENT lzf
         PARAMS { 4 261 8000000 }
      }
   }
   FILLVALUE {
      FILL_TIME H5D_FILL_TIME_IFSET
      VALUE  0
   }
   ALLOCATION_TIME {
      H5D_ALLOC_TIME_INCR
   }
   ATTRIBUTE "chunks" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
   }
}
}
```

crusaderky:
`coverage` shows that the whole compression support is severely lacking in test coverage.

ArvidJB:
[...]

PR #457, opened 2025-08-25
Filters overhaul; add shuffle, fletcher32, scaleoffset

crusaderky:
- Closes https://github.com/deshaw/versioned-hdf5/issues/449
- Rework the implementation of filters
- Add `shuffle`, `fletcher32`, and `scaleoffset` parameters to `create_dataset` and `modify_metadata`
- Fix bug where `modify_metadata` would revert `compression` and `compression_opts` to their default value when they are not explicitly listed. For example, `modify_metadata(ds, fillvalue=123)` would decompress a dataset! You now have to explicitly pass `modify_metadata(ds, compression=None)`.
- `.compression` and `.compression_opts`  properties now return the numerical IDs for custom compression filters (e.g. Blosc, Blosc2) in staged datasets. Note that this is unlike `h5py.Dataset.compression`, which incorrectly returns None.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -728,12 +728,7 @@ def test_resize_int_types_grow(tmp_path):
 
 def test_resize_sparse(vfile):
     with vfile.stage_version("version1") as group:
-        ds = group.create_dataset(
-            "data",
-            shape=(5, 5),
-            chunks=(3, 3),
-            maxshape=(None,),
-        )
+        ds = group.create_dataset("data", shape=(5, 5), chunks=(3, 3))
         ds[0, 0] = 1
         ds[4, 4] = 2
 
@@ -2978,8 +2973,10 @@ def test_other_compression_bad_value(vfile):
 
 
 @pytest.mark.parametrize("raw", [True, False])
-def test_other_compression_validates(h5file, raw):
-    """Test that other compression types validate correctly."""
+def test_blosc_compression_validates(h5file, raw):
+    """Test that third-party compression filters from hdf5plugin or pytables
+    such as Blosc validate correctly.
+    """
     hdf5plugin = pytest.importorskip("hdf5plugin")
 
     if raw:
@@ -3027,3 +3024,28 @@ def test_create_dataset_dtype_arg(vfile, dtype):
         dset = sv["y"]
         assert isinstance(dset.dtype, np.dtype)
         assert dset.dtype == np.dtype(dtype)
+
+
+@pytest.mark.parametrize(
+    "name,value",
+    [
+        ("fletcher32", True),
+        ("scaleoffset", 5),
+        ("shuffle", True),
+    ],
+)
+@pytest.mark.parametrize("sparse", [False, True])
+def test_create_dataset_other_filters(vfile, name, value, sparse):
+    kwargs = {"shape": (2,)} if sparse else {"data": [1, 2]}
+    kwargs[name] = value
+
+    with vfile.stage_version("r0") as sv:
+        x = sv.create_dataset("x", **kwargs)
+        assert getattr(x, name) == value
+
+    raw_data = vfile.f["_version_data"]["x"]["raw_data"]
+    assert getattr(raw_data, name) == value
+
+    with vfile.stage_version("r1") as sv:
+        x = sv["x"]
+        assert getattr(x, name) == value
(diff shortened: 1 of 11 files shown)


tests/test_api.py
```diff
@@ -2978,8 +2973,10 @@ def test_other_compression_bad_value(vfile):
 
 
 @pytest.mark.parametrize("raw", [True, False])
-def test_other_compression_validates(h5file, raw):
-    """Test that other compression types validate correctly."""
+def test_blosc_compression_validates(h5file, raw):
+    """Test that third-party compression filters from hdf5plugin or pytables
+    such as Blosc validate correctly.
+    """
```
crusaderky, line 2979: Now you can just run `pytest -k blosc` and hit all the relevant tests.

tests/test_backend.py
```diff
@@ -742,21 +743,26 @@ def test_write_dataset_compression(h5file):
     nchunks = int(np.ceil(data.size / chunksize))
 
     slices1 = write_dataset(
-        h5file, "test_data", data, compression="gzip", compression_opts=3
+        h5file,
+        "test_data",
+        data,
+        filters=Filters(compression="gzip", compression_opts=3),
```
crusaderky, line 749: This API is definitely less ergonomic than before. However, these are internal functions and this bit of ugliness is limited to a handful of unit tests.

tests/test_wrappers.py
```diff
@@ -306,32 +306,47 @@ def test_compression(vfile, group_name, name, sparse):
         assert ds.compression_opts == 5
 
 
-def test_compression_hotswap_to_inmemoryarraydataset(vfile):
-    """Test .compression and .compression_opts properties after DatasetWrapper hot-swaps
+def test_filters_hotswap_to_inmemoryarraydataset(vfile):
+    """Test .compression, .compression_opts, .shuffle, and .fletcher32 after
+    DatasetWrapper hot-swaps
```
crusaderky, line 311: can't test `scaleoffset` in the same test as it is mutually exclusive with `fletcher32`

versioned_hdf5/backend.py
```diff
@@ -68,17 +74,117 @@ def initialize(f):
...
+                        k = int(k)
+                    except ValueError:
+                        continue
+                    compression, compression_opts = k, v
+                    logging.warning(
+                        "No default compression detected in this dataset. "
+                        f"Guessed {compression=} {compression_opts=} "
+                        f"from {ds._filters=}."
+                    )
+                    break
```
crusaderky, line 157: This code path is untested. I'm unsure how to write a test for it.
The alternative was not to whitelist blosc etc. and just run this path. However, that carried the risk of encountering a custom non-compression filter plus blosc, which would mangle things up.

crusaderky:
Ready for review. Docs failure is fixed by #458.

versioned_hdf5/wrappers.py
```diff
@@ -221,21 +232,50 @@ def create_dataset(
...
         for k, v in kwds.items():
-            if v is not None and not (k == "maxshape" and all(i is None for i in v)):
+            if v is not None:
+                if k == "maxshape" and isinstance(v, tuple) and set(v) == {None}:
+                    continue
                 warnings.warn(
                     f"The {k} parameter is currently ignored for versioned datasets.",
                     stacklevel=2,
                 )
+        del kwds
```
peytondmurray, line 255: Why do we do this here? `kwds` is not used further down, right? I guess this just removes the temptation to use this later on?

peytondmurray (approved):
🚀

versioned_hdf5/backend.py
```diff
@@ -68,17 +74,117 @@ def initialize(f):
...
+                32015,  # Zstandard
+                32017,  # SZ
+                32024,  # SZ3
+                32018,  # FCIDECOMP
+                32028,  # SPERR
+            )
+
+            for filter_id in CUSTOM_COMPRESSION_FILTERS:
+                try:
+                    compression_opts = ds._filters[str(filter_id)]
```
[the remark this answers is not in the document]
crusaderky: Fairly sure that libhdf5 uses int IDs exclusively. No idea about h5py's design choice here.

versioned_hdf5/wrappers.py
crusaderky: Correct, it's just for code sanity.

Issue #466, opened 2025-12-04
h5py 3.15 broke Linux wheels


crusaderky:
Linux wheels for the latest released version of versioned-hdf5 (2.1.0) work against the wheels for h5py-3.14.0, and are broken against 3.15.0 and 3.15.1:

```bash
$ pip install h5py==3.14.0 versioned-hdf5==2.1.0
$ python -c 'import versioned_hdf5'  # OK
$ pip install h5py==3.15.0 versioned-hdf5==2.1.0
$ python -c 'import versioned_hdf5'
Traceback (most recent call last):
    from versioned_hdf5.slicetools import spaceid_to_slice
ImportError: libhdf5-8c29085d.so.310.5.1: cannot open shared object file: No such file or directory
```
The wheels CI job is also failing to produce new wheels with the same error message: https://github.com/deshaw/versioned-hdf5/actions/runs/19921831450

MacOSX wheels are unaffected.
Linux Python 3.9 wheels are unaffected because the last version of h5py supporting Python 3.9 is 3.14.
conda-forge is unaffected.

crusaderky:
The root cause is that
- h5py <=3.14.0 was built in `ghcr.io/h5py/manylinux2014_x86_64-hdf5`, whereas 
- h5py >=3.15.0 is built in `ghcr.io/h5py/manylinux_2_28_x86_64-hdf5`.

The two ship different versions of libhdf5. Same for aarch64.

As such,
- versioned-hdf5 built from sources (any platform) requires h5py >=3.8.0
- versioned-hdf5 conda-forge packages (any platform) requires h5py >=3.8.0
- versioned-hdf5 MacOSX wheels require h5py >=3.8.0
- versioned-hdf5==2.1.0 Linux x64 and ARM wheels require h5py wheels >=3.8.0,<3.15.0
- versioned-hdf5>=2.2.0 (next release) Linux x64 and ARM wheels will require h5py wheels >=3.15.0

Because Python 3.9 is stuck with h5py<=3.14.0, this makes it quite cumbersome to continue supporting Python 3.9.

PR #469, opened 2025-12-05
Require Python >=3.10; add 3.14 wheels

crusaderky:
As explained in #466, continued support for Python 3.9 has become unwieldy.
CI for wheels can be seen in #467.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -8,7 +8,7 @@ jobs:
         shell: bash -el {0}
     strategy:
       matrix:
-        python-version: ['3.13']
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -52,13 +52,13 @@ jobs:
         if: matrix.h5py-version == 'latest'
         run: |
           # Build against system-wide libhdf5
-          pip install h5py --no-binary :all:
+          pip install h5py --no-binary h5py
```
crusaderky: h5py has numpy as a build requirement. Don't compile it from sources.

PR #480, opened 2026-01-15
Fix numpy version parsing to account for local versions

ArvidJB:
If the numpy version contains a [local version](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers) the parsing logic breaks. E.g., the numpy version could be "2.3.0+deshaw3".

This change uses the official Python version regex to parse versions.

### DIFF
--- a/versioned_hdf5/tools.py
+++ b/versioned_hdf5/tools.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 from typing import Any
 
 import ndindex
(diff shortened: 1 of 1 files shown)


crusaderky:
This PR is technically correct, but I think it would be cleaner to just add a runtime dependency to `packaging` (it's already a test dependency). What do you think about it? It's almost certainly already pulled in by some other package anyway.
```python
>>> from packaging.version import Version
>>> Version("2.3.0+deshaw3").release
(2, 3, 0)
```

crusaderky:
Alternatively, we could use `np.lib.NumpyVersion` which is already used elsewhere:

```python
_v = np.lib.NumpyVersion("2.3.0+deshaw3")
(_v.major, _v.minor, _v.bugfix)
(2, 3, 0)
```

PR #495, opened 2026-03-26
Test vs. h5py 3.16 and hdf5 2.1

crusaderky:
Update h5py to 3.16. There are some minor differences re. hdf5plugin reporting.

Add test environment in CI with hdf5 2.1, using h5py built from sources.
This is a transitory situation. Once h5py upgrade their conda packages to pin hdf5 2.1, we'll flip our setup and have a dedicate `hdf5-114` environment instead.


peytondmurray (approved):
Thanks! 🚀

PR #496, opened 2026-03-26
Mismatched fixed-string dtypes updates in memory can swap dtype

crusaderky:
Fix regression where a ArrayDataset of fixed-size strings is incorrectly swapped with a different array of a differently sized strings.


peytondmurray (approved):
Let me check my understanding: h5py currently uses 'vlen' in the dtype metadata to indicate "variable-length strings" (NOT NpyString but rather python objects), whereas fixed-length strings are created with 'h5py_encoding' metadata. Is that right? See https://github.com/h5py/h5py/blob/d2f2ba86ed03091eab8985cc2bb110fa6630a77d/h5py/h5t.templ.pyx#L1897

If that's right, it seems like that `h5py_encoding` was indeed there by mistake.

Issue #477, opened 2026-01-12
Linux wheels CI broken; h5py 3.16 will break wheels compatibility again


crusaderky:
Our Linux wheels CI broke a few hours ago, due to the cibuildwheel images getting very recently upgraded (with the same name as before) to libhdf5 2.0.

This means that:
- We can't release a new versioned-hdf5 version with Linux wheels until h5py 3.16.0 comes out
- When it does come out, it will be pinned to depend on `h5py >=3.16.0`
- versioned-hdf5 2.2.1 Linux wheels are pinned against `h5py >=3.15.0`, but actually only work vs. `h5py >=3.15.0,<3.16.0`.
- This is just like versioned-hdf5 2.1.0 and older were pinned against `h5py >=3.8.0`, but the Linux wheels actually only work against `h5py >=3.8.0,<3.15.0` (#466)

MacOS wheels, conda packages, and source installation are unaffected.

A more in detail analysis and discussion upstream: https://github.com/h5py/hdf5-manylinux/issues/25

PR #497, opened 2026-03-26
Reinstate wheels (with exact h5py pins)

crusaderky:
XREF #477


peytondmurray (approved):
Since the cibw action version has been upgraded, I'll be curious to see if the PyPI upload works with the python version matrix upon next release. Anyway, thanks for dealing with this 🙏

PR #499, opened 2026-04-22
Build with meson 1.11

crusaderky:
Upgrading meson from 1.10 to 1.11 broke Linux CI, as meson has started calling /usr/bin/pkgconfig which in turn points to the system-wide Python interpreter.

This PR calls `pixi update`, but carefully avoids upgrading hdf5 from 1.14 to 2.1 (which is in scope for #498).


crusaderky:
FYI @peytondmurray

PR #500, opened 2026-04-22
Repoint manylinux images

### LINKED ISSUE #477: Linux wheels CI broken; h5py 3.16 will break wheels compatibility again

crusaderky:
Closes #477 

Prevent future changes to h5py/manylinux from breaking versioned-hdf5 linux wheels overnight.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,8 +151,12 @@ build-verbosity = 1
 build-frontend = { name = "pip", args = ["--only-binary", "numpy"] }
 # This must be kept in sync with h5py/pyproject.toml, as of the exact
 # H5PY_VERSION set in .github/workflows/wheels.yml.
-manylinux-x86_64-image = "ghcr.io/h5py/manylinux_2_28_x86_64-hdf5"
-manylinux-aarch64-image = "ghcr.io/h5py/manylinux_2_28_aarch64-hdf5"
+# Note that it points to the fork https://github.com/deshaw/hdf5-manylinux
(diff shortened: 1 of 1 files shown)


crusaderky:
Currently failing with

> docker image inspect ghcr.io/deshaw/manylinux_2_28_x86_64-hdf5 --format '{{.Os}}/{{.Architecture}}'
>  Error response from daemon: No such image: ghcr.io/deshaw/manylinux_2_28_x86_64-hdf5:latest
>  Error response from daemon: Head "https://ghcr.io/v2/deshaw/manylinux_2_28_x86_64-hdf5/manifests/latest": unauthorized

PR #502, opened 2026-04-30
Fix outstanding security issues with GH workflows

peytondmurray:
This PR fixes a number of issues with the CI configuration:

- Workflow-level permissions have been set to `none` for every category, meaning the `GITHUB_TOKEN` by default explicitly has no access to anything. This is fine for the majority of our workflows only need to checkout publicly accessible code (via `actions/checkout`), which doesn't require any permissions.
- For those jobs that require elevated permissions (`publish_docs.yml`, `pypi_publish.yml`), elevated permissions are _only_ set on jobs that require them, not all jobs (i.e. the entire workflow).
- We no longer persist `actions/checkout` credentials because we don't use git anywhere in these workflows.
- All action versions have been pinned to the hashes associated with the tags we were using before to prevent mutable-git-reference attacks.
- A few places where template injection was possible, I've moved the templates into environment variables (which are expanded by the shell, instead of in the action itself).

@crusaderky I used zizmor here to guide me on these fixes. We can make this a pre-commit hook or just use it in the lint action, if that makes sense to do so. Do you have a preference?

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,5 +1,6 @@
 name: Benchmarks
 on: [push, pull_request]
+permissions: {}
 jobs:
   benchmarks:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


crusaderky:
> @crusaderky I used zizmor here to guide me on these fixes. We can make this a pre-commit hook or just use it in the lint action, if that makes sense to do so. Do you have a preference?

I think just the lint action is good enough I think?

crusaderky (approved):

PR #503, opened 2026-05-18
Add AGENTS.md and AI development guidelines

crusaderky:
Add init file for all coding agents. 
Also include symlink CLAUDE.md -> AGENTS.md.

Add development guidelines for AI contributions.
Add skill for agents to open PRs unsupervised: draft state, and clearly recognisable as such.


CLAUDE.md
crusaderky, line 1: this is a symlink to AGENTS.md. (which is not great for windows claude users, but IMHO better than a manual copy)

.claude
crusaderky, line 1: this is a symlink to .agents (which is not great for windows claude users, but IMHO better than a manual copy)

ArvidJB:
Looks great, thanks for working on this! I think this is ready to be merged.

Issue #398, opened 2024-10-31
Redesign the commit stage


crusaderky:
Downstream of #386, the commit stage - when you exit the `stage_version(...)` context manager - is by far what takes the longest time. This is unsurprising, as #386 didn't touch it.

We should extend the redesign of #386 to everything that happens afterwards, which means hashing and writing to disk.
The StagedChangesArray introduced in #386 offers solid infrastructure to build on.

# Phases
This rewrite can happen in separate phases, each with an incremental benefit to production:
1. push into the StagedChangesArray the hashing phase and the writing to hdf5 when writing from a InMemoryDataset or InMemorySparseDataset, leaving the creation of the virtual dataset to the legacy code;
2. rewrite the creation of the virtual dataset downstream of InMemoryDataset or InMemorySparseDataset in a way that natively interfaces with the metadata of the StagedChangesArray, can be much more aggressively cythonized, and doesn't use the data_dict nor ndindex at any point;
3. rewrite create_dataset and all other machinery on top of the StagedChangesArray so that the legacy code can be retired.

# Performance expectations
Committing a 3 GiB dataset with 40,000x 80 kiB chunks (all chunks modified and different from the previous values) takes

- 12.7s today (as measured on NVMe Linux desktop)
- ~9.1s expected after phase 1
- ~5.8s expected after phase 2

A granular breakdown of these numbers appers later in this document.

# Low level design

**NOTE:** this section expects the reader to be already familiar with the low level design of #386. If you aren't, please stop here and read https://github.com/deshaw/versioned-hdf5/blob/17c687d3e1f1bfb64781a7b9039dbb621f5a46b4/docs/staged_changes.rst first.

Create a new `StagedChangesArray.hash_tables` list, matching the `slabs` list.
Every slab has a matching hash_table with dtype=hsize_t and shape=(n, 4), where n is the number of chunks in the slab (so every row can contain exactly one sha256 hash).

`hash_tables[i][j, :]` is the hash for the chunk on the slab at `slabs[i][j * chunk_size[0]:(j + 1) * chunk_size[0]]`.

New requirement: base slabs must be aligned to `chunk_sizes[0]`. This is de facto already true for the `raw_data` slab.

New hash tables are created with `np.zeros` together with the matching slabs. A hash set to 0 means that it's uninitialised.

The hash_tables entry for the raw_data base slab is an actual np.ndarray loaded from disk, not a h5py.Dataset.

### Committing changes
[...]

PR #505, opened 2026-05-26
read_many_slices NumPy->HDF5

crusaderky:
Towards #398
Add functionality to `read_many_slices` to write from NumPy o HDF5.
Add performance benchmarks for the function.


peytondmurray (approved):
Thanks for this!

PR #510, opened 2026-06-15
Fix wheels CI

crusaderky:
- Fix failure vs. pytest 9.1.0 (which is unpinned in wheels CI)
- Fix failure in Windows CI
```
  Generator

    Visual Studio 17 2022

  could not find any instance of Visual Studio.
```
- Fix very frequent random failure to download zlib sources
- Pin manylinux images to add resiliency from upstream rebuilds (there was a rebuild on June 18th). **Note:** the hashes are not pinned upstream, so they may get eventually garbage collected. This is not a robust fix.


crusaderky:
Bunch of failures during the test phase are fixed by https://github.com/deshaw/versioned-hdf5/pull/512
See https://github.com/deshaw/versioned-hdf5/pull/515 for cumulative run

crusaderky:
Great. This was green 3 days ago (https://github.com/deshaw/versioned-hdf5/actions/runs/27606710922).
Now hdf5-manylinux is broken again.

ImportError: libhdf5-4762e474.so.320.0.0: cannot open shared object file: No such file or directory

h5py hasn't changed in pypi (we pin it anyway).
There haven't been any commits to the hdf5-manylinux repo.

[EDIT] FIXED by pinning.
## Important note
Nothing is pinned upstream, so obsolete builds may get eventually garbage collected. This is not a robust fix.

.github/workflows/wheels.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - name: Run cibw_before_all_macos
         if: runner.os == 'macOS'
-        run: bash ./ci/cibw_before_all_macos.sh "${{ github.workspace }}"
+        run: ci/cibw_before_all_macos.sh "${{ github.workspace }}"
```
peytondmurray, line 66: Should this use `shell: bash` as well? Isn't the default zsh on macOS (which I guess shouldn't affect the runner's ability to execute this)?

Also, it seems to be nearly equivalent to what's here already, except we need different file permissions to ensure this works. What's the rationale for this change?

ci/get_hdf5_if_needed.sh
```diff
@@ -60,10 +60,10 @@ else
                 export MACOSX_DEPLOYMENT_TARGET="10.9"
             fi
 
-            pushd /tmp
-            curl -sLO https://zlib.net/fossils/zlib-$ZLIB_VERSION.tar.gz
-            tar xzf zlib-$ZLIB_VERSION.tar.gz
-            cd zlib-$ZLIB_VERSION
+            # zlib download is highly flaky; use a saved copy in the git repo
```
peytondmurray, line 63: Big +1 on this, Thank you!

peytondmurray (approved):
Thanks, this looks like it will improve CI reliability substantially 🚀

.github/workflows/wheels.yml
crusaderky: >  Isn't the default zsh on macOS

No idea

> What's the rationale for this change?

just out of symmetry with the rest of the changes in the file.
The necessary change is below in
```
        run: bash ./ci/triage_build.sh "$ARCH" "${{ github.event.pull_request.head.sha || github.sha }}" "$PYTHONVERSION"
```
which on windows was executed by pwsh, which failed to replace $PYTHONVERSION with its value, and then handed over to bash.

PR #514, opened 2026-06-19
Redesign the commit stage: `hash_slab`

crusaderky:
Towards https://github.com/deshaw/versioned-hdf5/issues/398

New heavily cythonized `hash_slab` function, which replicates the design of `read_many_slices`.
This will eventualy replace HashTable. For now, the two exist side by side, producing identical outputs.

The new function is not plugged into anything yet. This will be the target of a follow-up PR.

# Dependency on OpenSSL
The ugly bit of this PR is that it introduces a hard dependency on OpenSSL.
This was unavoidable. A _much_ cleaner alternative was to just vendor a tiny module that does the sha256 calculation, but it turned out to be 10x slower than python's hashlib for large chunks.
Python's hashlib in turn vendors its own sha code, and to the best of my understanding it is not possible to tap into it without passing through the Python API.

The dependency on OpenSSL in turn requires a bunch of hacks to make it work with wheels - very painful, as it is usual for wheels.
For conda, this is a non-issue (just need to add a runtime dependency to the openssl package, which is already pulled in by `python`).
Compiling from sources, this should be a non-issue short of fairly exotic barebones Docker images.


versioned_hdf5/hash.pyx
```diff
@@ -0,0 +1,305 @@
...
+        _hash_shape(ctx, &shape[0], ndim)
+        if EVP_DigestFinal_ex(ctx, <unsigned char*>out, NULL) != 1:
+            return -1
+    finally:
+        EVP_MD_CTX_free(ctx)
+
+    return 0
+
+
+cdef void _hash_object_chunk(
```
peytondmurray, line 258: Looking at the old hashing function we had and remembering from all the ways string hashing was previously gotten wrong, I think this looks good. 👍

peytondmurray (approved):
Looks good. Did you end up running the benchmarks yet?

crusaderky:
> Looks good. Did you end up running the benchmarks yet?

yes, they are in the CI run:
https://github.com/deshaw/versioned-hdf5/actions/runs/28167966796/job/83424612115

you can see below how:

- performance is identical to pure-python for 8MB chunks
- at 64 kiB chunk size, overhead in Cython causd performance to degrade 2%; pure Python degraded 11% (Cython ix 1.1x faster than Python)
- at 1 kiB chunk size, cython performance degraded 45%; pure Python degraded 86% (Cython is 3.9x faster than Python)
- vstrings are between 1.7x and 3.5x faster depending on chunk size and string length
- NpyStrings are slightly _slower_ than object strings as expected (I have not implemented native hashing yet)

```

[ 4.84%] ··· hash.TimeHashSlab.time_hash_slab                                ok
[ 4.84%] ··· ============== ========== ==========
             --                      edge        
             -------------- ---------------------
               chunk_size     False       True   
             ============== ========== ==========
                (8, 16)      19.5±0ms   20.8±0ms 
               (64, 128)     10.9±0ms   11.0±0ms 
              (1024, 1024)   10.7±0ms   10.8±0ms 
             ============== ========== ==========

[ 6.45%] ··· hash.TimeHashSlab.time_hash_slab_naive                          ok
[ 6.45%] ··· ============== ========== ==========
             --                      edge        
             -------------- ---------------------
               chunk_size     False       True   
             ============== ========== ==========
                (8, 16)      75.7±0ms   83.5±0ms 
               (64, 128)     12.0±0ms   12.7±0ms 
              (1024, 1024)   10.7±0ms   14.2±0ms 
             ============== ========== ==========

[ 8.06%] ··· hash.TimeHashSlabStrings.time_hash_slab                         ok
[ 8.06%] ··· ======= ============ ========== ===========
             --                         chunk_size      
             -------------------- ----------------------
              dtype   max_nchars   (8, 16)    (64, 128) 
             ======= ============ ========== ===========
                O         8        2.87±0ms    2.51±0ms 
                O         64       4.21±0ms    3.87±0ms 
                O        256       8.32±0ms    7.98±0ms 
                T         8        4.33±0ms    3.39±0ms 
                T         64       5.56±0ms    5.18±0ms 
                T        256       9.95±0ms    10.8±0ms 
             ======= ============ ========== ===========

[ 9.68%] ··· hash.TimeHashSlabStrings.time_hash_slab_naive                   ok
[ 9.68%] ··· ======= ============ ========== ===========
             --                         chunk_size      
             -------------------- ----------------------
              dtype   max_nchars   (8, 16)    (64, 128) 
             ======= ============ ========== ===========
                O         8        9.91±0ms    8.48±0ms 
                O         64       10.7±0ms    9.68±0ms 
                O        256       15.6±0ms    13.9±0ms 
                T         8        11.1±0ms    9.34±0ms 
                T         64       12.7±0ms    11.0±0ms 
                T        256       17.1±0ms    15.6±0ms 
             ======= ============ ========== ===========
```

PR #516, opened 2026-06-24
Windows wheels: retry choco install

crusaderky:


peytondmurray (approved):
I guess the real test is to see if this passes. Thanks for this!

PR #517, opened 2026-06-25
Work around Cython 3.2.6 regression: `@with_phil` breaks function annotations

crusaderky:
## Problem

The `h5py-dev` CI job started failing at import time with:

```
TypeError: InMemoryDataset.__getitem__.__annotate__() missing 1 required positional argument: 'format'
```

## Root Cause

- https://github.com/cython/cython/issues/7767

which in turn broke `@with_phil`.

**Note:** only `h5py-dev` and `mindeps` compile h5py from sources.
`mindeps` uses an exact pin to the lowest supported version of cython.
all other envs use h5py from conda-forge, which was compiled many weeks ago, which in turn means an older version of cython was used to compile it.


versioned_hdf5/wrappers.py
```diff
@@ -703,18 +703,7 @@ def resize(self, size, axis=None):
...
-        args: (
-            str
-            | slice
-            | tuple
-            | list[int]
-            | list[bool]
-            | np.ndarray
-            | h5r.RegionReference
-        ),
-    ) -> np.ndarray:
```
crusaderky, line 717: this brute-force approach brings `__getitem__` in line with `__setitem__`.
IMHO this is cleaner than pinning an older cython version.

PR #518, opened 2026-07-01
CI refresh

crusaderky:
Notable: 
- pixi action no longer has a cleanup phase, resulting in slightly faster CI
- hdf5plugin 6.0.0 ->  7.0.0

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -31,9 +31,9 @@ jobs:
           fetch-tags: true
           persist-credentials: false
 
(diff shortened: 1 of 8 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -31,9 +31,9 @@ jobs:
           fetch-tags: true
           persist-credentials: false
 
-      - uses: prefix-dev/setup-pixi@1b2de7f3351f171c8b4dfeb558c639cb58ed4ec0 # v0
```
crusaderky, line 34: 0.9.6 -> 0.10.0

PR #521, opened 2026-07-14
Redesign the commit stage: `StagedChangesArray.commit()`

crusaderky:
- Towards #398
- Blocked by and incorporates #525

This PR integrates the hashing system of #514 into StagedChangesArray, which exposes a new `commit(empty=np.empty)` method.

## Brief workflow summary
A less hermetic explanation is available in the documentation changes in this same PR.

The following happens when you call `StagedChangesArray.commit()`:

1. StagedChangesArray generates a `HashPlan` for all staged slabs
2. then it uses the HashPlan to call `slicetools.hash_slab` once per each slab (added in #514)
3. that in turn populates the new `StagedChangesArray.hash_tables` list that runs parallel to `StagedChangesArray.slabs`
4. back to StagedChangesArray, which generates a `CommitPlan`
5. `CommitPlan` ingests all the hashes of the base slabs (loaded from hdf5) plus the newly-generated hashes of the staged chunks
6. `CommitPlan` deduplicates the chunks (full slab wins over base slab that wins over staged slab) using a C++ `std::unordered_map`
7. `CommitPlan` generates a plan to transfer all staged chunks that survived deduplication to a new base slab and generate a new hash_table with the hashes of all the chunks it planned to copy
8. back to `StagedChangesArray`, which invokes the `empty` callback to generate the new base slab. In unit tests, this is `np.empty`. After integration, it will be a writeable view of the h5py `raw_data` dataset, which has just been enlarged to accommodate the new chunks.
9. `StagedChangesArray` invokes `slicestools.read_many_slices` from the staged slabs (np.ndarray) to the base slab, which post integration will be a HDF5 dataset (numpy->hdf5 write added in #505)
10. You now have the following:
  - a new base slab with the new chunks, which is a view to the tail of the `raw_data` HDF5 dataset on disk;
  - a new hash table, which the wrapper just needs to append to the hash table hdf5 dataset on disk;
  - the `StagedChangesArray.changes()` method that returns only slices to either the old or the new base slab; this is transitorily still needed for as long as we use the legacy pure-python function to generate the virtual dataset
  - the `StagedChangesArray.slab_offsets` and `StagedChangesArray.slab_indices` arrays are ready to be trivially iterated upon by the future, heavily cythonized, function that will generate the VDS replacement (stage 2 of #398)

## Out of scope
This PR does NOT trigger `StagedChangesArray.commit()` when a staged dataset leaves its context manager. 
User-facing performance is unaltered as none of the new functionality is actually triggered by final users yet.
Integration will be in scope for one more PR downstream of this.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,6 +20,17 @@ author = "Quansight"
 # with the -W flag in the Makefile.
 nitpicky = True
 
+# These internal modules are not documented via autodoc, so Sphinx can't
+# resolve cross-references to them. Ignore them in nitpicky mode.
+nitpick_ignore = [
+    ("py:class", "versioned_hdf5.hashtable.Hashtable"),
+    ("py:meth", "versioned_hdf5.staged_changes.StagedChangesArray.commit"),
+    ("py:meth", "versioned_hdf5.hashtable.Hashtable.hash"),
+    ("py:mod", "versioned_hdf5.hash"),
+    ("py:mod", "versioned_hdf5.replay"),
+    ("py:func", "versioned_hdf5.backend.commit_staged_changes"),
+]
+
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,17 +20,6 @@ author = "Quansight"
 # with the -W flag in the Makefile.
 nitpicky = True
 
-# These internal modules are not documented via autodoc, so Sphinx can't
-# resolve cross-references to them. Ignore them in nitpicky mode.
-nitpick_ignore = [
-    ("py:class", "versioned_hdf5.hashtable.Hashtable"),
-    ("py:meth", "versioned_hdf5.staged_changes.StagedChangesArray.commit"),
-    ("py:meth", "versioned_hdf5.hashtable.Hashtable.hash"),
-    ("py:mod", "versioned_hdf5.hash"),
-    ("py:mod", "versioned_hdf5.replay"),
-    ("py:func", "versioned_hdf5.backend.commit_staged_changes"),
-]
-
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
--- a/versioned_hdf5/staged_changes.py
+++ b/versioned_hdf5/staged_changes.py
@@ -17,6 +17,7 @@ from cython import bint, ssize_t
 from numpy.typing import ArrayLike, DTypeLike, NDArray
 
 from versioned_hdf5.cytools import ceil_a_over_b, count2stop, np_hsize_t
+from versioned_hdf5.hash import hash_slab
 from versioned_hdf5.slicetools import read_many_slices
 from versioned_hdf5.subchunk_map import (
     EntireChunksMapper,
@@ -170,6 +171,24 @@ class StagedChangesArray(MutableMapping[Any, T]):
     #: slabs[0] is never dereferenced.
     slabs: list[NDArray[T] | None]
 
+    #: List parallel to :attr:`slabs` (same length, None wherever ``slabs`` is None).
+    #: An uncomputed hash table is also set to None.
+    #:
+    #: ``hash_tables[i]`` is a C-contiguous array of dtype ``uint64`` and shape
+    #: ``(n_i, 4)``, where ``n_i`` is the number of chunks in ``slabs[i]``,
+    #: and each row holds one SHA256 digest (4 * 8 = 32 bytes).
+    #:
+    #: ``hash_tables[i][j]`` is the hash of the chunk at
+    #: ``slabs[i][j*chunk_size[0]:(j+1)*chunk_size[0]]``
+    #: ignoring slab contents beyond the edge in case of edge chunks smaller
+    #: than chunk_size.
+    #:
+    #: A row set to all-zeros means the hash is uninitialized. This happens when
+    #: a staged chunk is created and then deleted, but other chunks on the slab remain.
+    #:
+    #: base slab hashes are typically loaded from disk and passed to ``__init__``.
+    hash_tables: list[NDArray[np.uint64] | None]
+
     #: Number of base slabs in the slabs list.
     #: Staged slabs start at index n_base_slabs + 1.
     n_base_slabs: int
@@ -187,6 +206,7 @@ class StagedChangesArray(MutableMapping[Any, T]):
         slab_indices: ArrayLike,
         slab_offsets: ArrayLike,
         fill_value: Any | None = None,
+        base_hash_tables: Sequence[NDArray[np.uint64]] | None = None,
     ):
         # Sanitize input (e.g. convert np.int64 to int)
         shape = tuple(int(i) for i in shape)
@@ -231,6 +251,23 @@ class StagedChangesArray(MutableMapping[Any, T]):
         self.slabs.extend(base_slabs)
         self.n_base_slabs = len(base_slabs)
 
+        # Build the hash_tables list, parallel to slabs.
+        self.hash_tables = [None]  # Full slab
+
+        if base_hash_tables is None:
+            # This only happens in unit tests
+            self.hash_tables += [None] * len(base_slabs)
+        else:
+            for base_slab, hash_table in zip(base_slabs, base_hash_tables, strict=True):
+                n_chunks_i = int(ceil_a_over_b(base_slab.shape[0], chunk_size[0]))
+                hash_table_shape = (n_chunks_i, 4)
+                if hash_table.shape != hash_table_shape:
+                    raise ValueError(
+                        f"{hash_table.shape=}; expected {hash_table_shape}"
+                    )
+                bht = np.ascontiguousarray(np.asarray(hash_table, dtype=np.uint64))
+                self.hash_tables.append(bht)
+
         self.slab_indices = asarray(slab_indices, dtype=np_hsize_t)
         self.slab_offsets = asarray(slab_offsets, dtype=np_hsize_t)
 
@@ -243,6 +280,13 @@ class StagedChangesArray(MutableMapping[Any, T]):
     @property
     def full_slab(self) -> NDArray[T]:
         return cast(NDArray[T], self.slabs[0])
+
+    @property
+    def full_hash_table(self) -> NDArray[T] | None:
+        """The hash_table for the full slab is calculated lazily only when actually
+        needed to deduplicate the staged chunks
+        """
+        return self.hash_tables[0]
 
     @property
     def fill_value(self) -> NDArray[T]:
@@ -309,8 +353,16 @@ class StagedChangesArray(MutableMapping[Any, T]):
         return self.slabs[1 : self.staged_slabs_start]
 
     @property
+    def base_hash_tables(self) -> Sequence[NDArray[np.uint64] | None]:
+        return self.hash_tables[1 : self.staged_slabs_start]
+
+    @property
     def staged_slabs(self) -> Sequence[NDArray[T] | None]:
         return self.slabs[self.staged_slabs_start :]
+
+    @property
+    def staged_hash_tables(self) -> Sequence[NDArray[np.uint64] | None]:
+        return self.hash_tables[self.staged_slabs_start :]
 
     @property
     def has_changes(self) -> bool:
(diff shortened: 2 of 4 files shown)


versioned_hdf5/_commit_hash.hpp
crusaderky, line 1: I hate cython. The whole hashmap system would have been 20 lines with Rust+PyO3.

pixi.toml
```diff
@@ -30,11 +30,13 @@ patchelf = "*"
 
 [feature.build.target.win.dependencies]
 clang = "*"
+clangxx = "*" # staged_changes.pyx is compiled as C++
```
crusaderky: no idea why this is not necessary on macosx.

crusaderky:
@krassowski this has been rebased on top of master

crusaderky:
**Note:** under the current design, the commit time increases with the number of obsolete versions.

This costs, [as measured here](https://github.com/deshaw/versioned-hdf5/pull/525#issuecomment-5137282038),  ~142 ns per obsolete chunk, and the opportunity to skip rewriting even a single a chunk IMHO justifies the cost of parsing all obsolete hashes every time. 

It would be very complicated to implement a bounding policy e.g. "only keep obsolete chunks from the previous 2 versions" because such chunks could start literally anywhere in the past, so you'd need to either retrieve the previous virtual datasets or duplicate their information somewhere, and both operations would be very expensive.

docs/conf.py
```diff
@@ -20,6 +20,17 @@
...
+# These internal modules are not documented via autodoc, so Sphinx can't
+# resolve cross-references to them. Ignore them in nitpicky mode.
+nitpick_ignore = [
+    ("py:class", "versioned_hdf5.hashtable.Hashtable"),
+    ("py:meth", "versioned_hdf5.staged_changes.StagedChangesArray.commit"),
+    ("py:meth", "versioned_hdf5.hashtable.Hashtable.hash"),
+    ("py:mod", "versioned_hdf5.hash"),
+    ("py:mod", "versioned_hdf5.replay"),
+    ("py:func", "versioned_hdf5.backend.commit_staged_changes"),
+]
```
krassowski: Is this needed? They don't seem referenced? On the other hand, should `StagedChangesArray.commit` land in the docs?

krassowski (commented):
Should `commit` be atomic or at least tolerant to exceptions thrown and clear hashes in such a case? See clauded reproducer below.

<details>

```py

GEOM = {"shape": (2,), "chunk_size": (2,)}


def enospc(*_args, **_kwargs):
    """Stand-in for the `empty` callback failing partway through commit().

    In production `empty` enlarges the raw_data hdf5 dataset, so this is what a full
    disk looks like. A Ctrl-C landing inside commit() leaves the same state and needs
    no callback at all.
    """
    raise OSError(28, "No space left on device")


# --- Version 1: commit() is interrupted after hashing; the user carries on.
v1 = StagedChangesArray.full(dtype="i8", **GEOM)
v1[:] = 1
try:
    v1.commit(empty=enospc)  # _calc_hashes() has already run
except OSError:
    pass
v1[:] = 2  # the staged slab is overwritten; its hash is not invalidated
v1.commit()  # _calc_hashes() skips the slab, so hash(1) is committed for data 2

raw_data = np.asarray(v1.base_slabs[-1])
hash_table = np.asarray(v1.base_hash_tables[-1])
print(f"v1 stored {raw_data.tolist()} and recorded the hash of [1, 1]")

# --- Version 2 opens v1's raw_data + hash table, as InMemoryDataset will.
v2 = StagedChangesArray(
    base_slabs=[raw_data],
    base_hash_tables=[hash_table],
    slab_indices=[1],
    slab_offsets=[0],
    **GEOM,
)
v2[:] = 1
v2.commit()  # hash([1, 1]) matches the poisoned row -> deduplicated away

print(f"v2 wrote [1, 1] and reads back {np.asarray(v2).tolist()}")
assert v2.n_base_slabs == 1, "v2 wrote a new chunk"
assert (np.asarray(v2) == 1).all(), "CORRUPTION: v2's data was never stored"
```

</details>

crusaderky:
> Should commit be atomic or at least tolerant to exceptions thrown and clear hashes in such a case? 

The write to the new base slab can totally fail halfway through. On-disk integrity is catered for by the next PR.

As a rule of thumb, `_apply_mutating_plan` doesn't cope with SIGINT or other exceptions (like disk full). It's not just a matter of wrapping everything in a try...except, because the plans write back to `slab_indices` and `slab_offsets`. In other words, after you *formulate* a MutatingPlan, *you must enact it* otherwise the whole StagedChangesArray will be corrupted. There is an option to make it resilient to the plan being aborted: the `copy=True` flag, e.g. in `_commit_plan`. This flag is True in some unit tests and False everywhere else, because it costs a deep copy of `slab_indices` and `slab_offsets` every time a MutatingPlan is formulated. This becomes unacceptable eg when the user calls `__setitem__` many times in a tight for loop: given small enough chunks, with copy=True, `__setitem__` would change 
- from O(number of modified chunks) 
- to O(total number of chunks).

crusaderky:
I am fairly unconcerned about StagedChangesArray robustness vs. SIGINT and OSError.

While in a vacuum it looks like a serious design flaw, if you zoom out you realise that there is no public-facing commit() method: the only way for a user to write to disk is to exit the `stage_version(...)` context manager.
This makes it impossible to catch-and-retry a failed StagedChangesArray.commit() wiithout explicitly calling `ctx = stage_version(...).__enter__()` ; alternatively it would take a `try...except Exception` around `__setitem__` (that blindly catches OSError and KeyboardInterrupt) to put oneself at risk of disk corruption.

To recap:
THIS can result in a partial write to disk e.g. on disk full, which is robustly handled:
```python
with vf.stage_version(...) as v:
     v["ds_name"] = ...
```

While these are the only ways one could conceivably cause a disk corruption:
```python
with vf.stage_version(...) as v:
     try:
        v["ds_name"] = ...  # Partially updating chunks triggers disk READ
    except Exception:  # Disk read could fail (eg unresponsive NFS server); alternatively there could be KeyboardInterrupt
        pass
```

```python
sv = vf.stage_version(...)
v = sv.__enter__()
v["ds_name"] = ...
for retry in range(3):
    try:
        sv.__exit__()
        break
    except Exception:
        if retry == 2:
            raise
```
Both are textbook antipatterns on behalf of the user.

docs/conf.py
crusaderky: sphinx build used to fail without these. Looks they are no longer relevant to the latest version of staged_changes.rst. Removed.

crusaderky:
@krassowski I've addressed all comments

versioned_hdf5/staged_changes.py
```diff
@@ -244,6 +281,13 @@ def __init__(
     def full_slab(self) -> NDArray[T]:
         return cast(NDArray[T], self.slabs[0])
 
+    @property
+    def full_hash_table(self) -> NDArray[T] | None:
```
krassowski: Should the return type be `np.uint64` as in other properties that expose fragments of `hash_tables`?

```suggestion
    def full_hash_table(self) -> NDArray[np.uint64] | None:
```

versioned_hdf5/staged_changes.py
```diff
@@ -1734,6 +1946,336 @@ def head(self) -> str:
...
+        chunk_size_0: hsize_t = chunk_size[0]
+
+        _, mappers = index_chunk_mappers((), shape, chunk_size)
+        if not mappers:  # size-0 array: no chunks at all
+            return
+
+        if slab_idx_filter[0]:
+            # Hash full slab. Do not pass it to _chunk_in_selection below
+            # as it would return one row for each chunk where it is used.
+            slab_idx_filter[0] = False
```
krassowski: Since this argument gets mutated, constructing `HashPlan ` twice with the same arguments would lead to different results. This is inconsequential right now because the `slab_idx_filter` is created just before `HashPlan` gets instantiated. I don't know if the cost of copying it would be something to worry about. This could be ignored if we treat `HashPlan` as internal API only or addressed by mentioning it in the docstring.

versioned_hdf5/staged_changes.py
```diff
@@ -1734,6 +1946,336 @@ def head(self) -> str:
         return "ResizePlan<" + super().head
 
 
+@cython.cclass
+@dataclass(repr=False, eq=False)
```
krassowski: Nice to have: since the code already uses keyword-only constructors, why not enforce it here and for the other data classes? Unless there is a reason not to do so.

```suggestion
@dataclass(repr=False, eq=False, kw_only=True)
```

versioned_hdf5/staged_changes.py
```diff
@@ -2255,6 +2254,23 @@ def __init__(
...
+        for dim in range(ndim):
+            last_chunk_idx = self.slab_indices.shape[dim] - 1
+            trim = shape[dim] - last_chunk_idx * chunk_size[dim]
+            if trim == chunk_size[dim]:
+                continue  # No edge chunks along this axis
+            edge = (slice(None),) * dim + (slice(last_chunk_idx, None),)
+            edge_offsets = self.slab_offsets[edge][
+                self.slab_indices[edge] == new_slab_idx
+            ]
+            tplan_count[edge_offsets // cs0, dim] = trim
```
krassowski, line 2271: Is it easy to add a test for this?

krassowski (approved):
This looks good to merge. I left some minor nitpicks/suggestions.

For completeness, noting that `refill` and `astype` have similar cache invalidation susceptibility as `__setitem__` had:

<details>

```python
import numpy as np

from versioned_hdf5.staged_changes import StagedChangesArray

GEOM = {"shape": (2,), "chunk_size": (2,)}


def enospc(*_args, **_kwargs):
    """A commit() that fails after hashing, e.g. raw_data hitting a full disk."""
    raise OSError(28, "No space left on device")


def part1_invariant() -> None:
    """Minimal: the hash survives an operation that changed the data."""
    for name, op in [
        ("refill(9)", lambda a: a.refill(9)),
        ("astype('f8')", lambda a: a.astype(np.dtype("f8"))),
    ]:
        a = StagedChangesArray.full(dtype="i8", fill_value=0, **GEOM)
        a[:] = [1, 0]
        a._calc_hashes()
        b = op(a)
        stale = b.hash_tables[1] is not None
        print(f"  {name:<13} stale hash survived: {stale}")


def part2_corruption() -> None:
    """The stale hash reaches the on-disk hash table and corrupts a later version."""
    v1 = StagedChangesArray.full(dtype="i8", fill_value=0, **GEOM)
    v1[:] = [1, 0]
    try:
        v1.commit(empty=enospc)  # hashes [1, 0], then fails
    except OSError:
        pass

    v2 = v1.refill(9)  # rewrites the slab to [1, 9]; hash not cleared
    v2.commit()  # records hash([1, 0]) for data [1, 9]
    print(f"  v2 committed {np.asarray(v2).tolist()} under the digest of [1, 0]")

    # A later version legitimately writes [1, 0] - the data that digest belongs to.
    v3 = StagedChangesArray(
        base_slabs=[np.asarray(v2.base_slabs[-1])],
        base_hash_tables=[np.asarray(v2.base_hash_tables[-1])],
        slab_indices=[1],
        slab_offsets=[0],
        fill_value=0,
        **GEOM,
    )
    v3[:] = [1, 0]
    v3.commit()  # deduplicated onto the chunk holding [1, 9]
    got = np.asarray(v3).tolist()
    print(f"  v3 wrote [1, 0] and reads back {got}")
    assert got == [1, 0], "CORRUPTION: v3's data was never stored"


if __name__ == "__main__":
    print("1. invariant")
    part1_invariant()
    print("2. corruption")
    part2_corruption()
    print("OK: no corruption")
```

</details>

But I understand that these are unreachable via public API/correct usage.

I see that `astype` and `refill` are mentioned in "Out of scope bonus: cache on read" in https://github.com/deshaw/versioned-hdf5/issues/398. I think that once that is implemented this edge case would go away?

> Both are textbook antipatterns on behalf of the user.

The second is an obvious anti-pattern. The first... Yes, but I could imagine someone arriving at a code which does that if they aimed to catch a different exception in a messy code. Maybe the right solution here is just to document the rollback guarantee of `stage_version`, and what are its limitations (i.e. if someone silently catches all `Exception`s or even worse `BaseException`s inside the context manager they ask for trouble).

crusaderky:
> I see that astype and refill are mentioned in "Out of scope bonus: cache on read" in https://github.com/deshaw/versioned-hdf5/issues/398.

I have second thoughts about cache-on-read. 
- it would change GetItemPlan from read-only and idempotent to a MutatingPlan, which in turn means that, when implementing free-threading, you need a per-dataset lock around all methods while now a rwlock is enough (allowing for parallel reads from memory, but not from the base slabs because there you get blocked by h5py's interpreter-wide lock).
- it would consume extra RAM for all users, to improve the specific use case that a user reads from the same chunks twice and is not already caching on their side
- it is redundant with HDF5's cache system, which by default is tiny but can be increased

versioned_hdf5/staged_changes.py
crusaderky: The whole module is internal API. 
Changed the parameter so that the issue doesn't arise, without the need for a copy.

versioned_hdf5/staged_changes.py
crusaderky: added test

krassowski:
> The whole module is internal API.

I know. I am not sure if users will know. We can hope they will know, but we can't be sure. I can imagine an AI agent trying to work around something or reuse code and going to import one of these classes, which we will then break on patch release if internals get changed. Maybe before release we could harden this by using `__all__` in `staged_changes.py` so that classes which should not be imported out of this module are not easy to import? Another approach would be renaming `staged_changes.py` to `_staged_changes.py`.

krassowski:
Thanks for addressing the suggestions! It's nice that `slab_idx_filter` got solved without an extra copy.

PR #525, opened 2026-07-21
Redesign the commit stage: StagedChangesArray.commit() C++ groundwork

crusaderky:
Towards #398

Write two C++ accelerated hash maps:

- **key:** sha256 of a chunk, **value:** (slab index, slab offset) pair
- **key:** (slab index, slab offset) pair, **value:** (slab index, slab offset) pair

These will be used by StagedChangesArray in #521 to deduplicate the chunks:
1. perform a full scan of all hashes present and past; store where each hash maps to on the live array. Skip those that have been deleted.
2. for each staged chunk, find the coordinates where it remaps to

These two hashmaps are unused so far. They are in a separate PR only for the sake of easing code review.
Linter error is transitory - a necessary evil to demonstrate that everything builds correctly.

P.S. I hate cython. The whole system would have been 20 lines with Rust+PyO3.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -30,11 +30,13 @@ patchelf = "*"
 
 [feature.build.target.win.dependencies]
 clang = "*"
+clangxx = "*" # staged_changes.pyx is compiled as C++
 llvm-tools = "*"
 pkg-config = "*" # Needed to find hdf5
 
 [feature.build.target.win.activation.env]
 CC = "clang"
+CXX = "clang++"
 LD = "llvm-ar"
 
 [feature.build.tasks]
@@ -99,8 +101,12 @@ clang = "=14"
 # hdf5 1.12 shows double-linking issues at runtime
 hdf5 = "=1.14.0"
 h5py = "=3.9.0"
-clang = "=14"
-llvm-tools = "=14"
+# This is for compatibility with windows-latest (windows-2025 at the moment of writing).
+# It may break when windows-latest is upgraded in the future to ship with a newer
+# version of MSVC.
+clang = "=20"
+clangxx = "=20"
+llvm-tools = "=20"
 
 [feature.hdf5-112]
 # hdf5 1.12.2 is the minimum version supported by h5py binaries in conda-forge.
--- a/versioned_hdf5/_commit_hash.hpp
+++ b/versioned_hdf5/_commit_hash.hpp
@@ -1,0 +1,69 @@
+// Custom hash functors and key types for the C++ std::unordered_map used by
+// staged_changes.CommitPlan.
+//
+// C++ std::hash has no specialization for std::array, std::pair, or std::tuple,
+// so we provide trivial custom hashers here. This keeps the Cython side free of
+// C++ boilerplate: staged_changes only has to declare the instantiated
+// unordered_map typedefs.
+//
+// This header has no dependencies beyond the C++ standard library.
+
+#pragma once
+
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <utility>
+
+namespace versioned_hdf5 {
+
+// Key for the {sha256 -> (slab_idx, slab_offset)} map.
+// A SHA256 digest is 4 x uint64 = 32 bytes. The digest is already a
+// cryptographically strong, uniformly distributed hash, so we can feed any one
+// of its words straight into std::hash<uint64_t> instead of hashing all four.
+struct ChunkHash {
+    std::array<uint64_t, 4> words;
+
+    ChunkHash() = default;
+    ChunkHash(uint64_t h0, uint64_t h1, uint64_t h2, uint64_t h3) noexcept {
+        words[0] = h0;
+        words[1] = h1;
+        words[2] = h2;
+        words[3] = h3;
+    }
+
+    bool operator==(const ChunkHash& other) const noexcept {
+        return words == other.words;
+    }
+};
+
+struct ChunkHashHash {
+    std::size_t operator()(const ChunkHash& key) const noexcept {
+        // words[0] is uniformly distributed (it's a SHA256 word), so a single
+        // pass through std::hash<uint64_t> is sufficient and collision resistant.
+        return std::hash<uint64_t>{}(key.words[0]);
+    }
+};
+
+// Key for the {(slab_idx, slab_offset) -> (slab_idx, slab_offset)} map.
+struct ChunkLoc {
+    uint64_t slab_idx;
+    uint64_t slab_offset;
+
+    ChunkLoc() = default;
+    ChunkLoc(uint64_t idx, uint64_t off) noexcept
+        : slab_idx(idx), slab_offset(off) {}
+
+    bool operator==(const ChunkLoc& other) const noexcept {
+        return slab_idx == other.slab_idx && slab_offset == other.slab_offset;
+    }
+};
+
+struct ChunkLocHash {
+    std::size_t operator()(const ChunkLoc& key) const noexcept {
+        std::size_t h = std::hash<uint64_t>{}(key.slab_offset);
+        return h ^ (std::hash<uint64_t>{}(key.slab_idx) << 1u);
+    }
+};
+
+}  // namespace versioned_hdf5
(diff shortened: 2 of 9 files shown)


pixi.toml
```diff
@@ -30,11 +30,13 @@ patchelf = "*"
 
 [feature.build.target.win.dependencies]
 clang = "*"
+clangxx = "*" # staged_changes.pyx is compiled as C++
```
crusaderky, line 33: no idea why this is not necessary on macosx.

krassowski (commented):
Questions I asked myself when looking at this PR:
- why is cythonization optional? This is not a comment on this PR as that is pre-existing, it just feels like it adds a bit of overhead with two implementations one needs to write and then review, and users could see much worse performance if they used plain Python one?
- what is the performance benefit of using unordered_map from C++ over dict? Do you have benchmarks for these two variants? Is this on performance critical path?
- should there be any tests added for the `_commit_hash` implementations?

crusaderky:
> * why is cythonization optional? This is not a comment on this PR as that is pre-existing, it just feels like it adds a bit of overhead with two implementations one needs to write and then review, and users could see much worse performance if they used plain Python one?

Oh yes, performance will be awful. Final users are not meant to disable cythonization.
Ensuring that _most_ of the cython modules can be executed as pure-python is a workaround to Cython's exceptionally poor tooling. Switching to pure-pyhon allows a developer to trivially perform coverage analysis, line-by-line debugging,  skip recompilation after trivial changes, use all sort of static code checkers (both command-line and integrated in your IDE of choice), and generate Sphinx docs.

_commit_hash is a special case that requires a dev-only pure-python stub, because it's a pure C++ header which pure python can't import.

> * what is the performance benefit of using unordered_map from C++ over dict? 

It skips the entire Python API. Instead of iterating on PyObject tuples of PyObject ints or structs, hashing them, increasing their refcount, and then storing them into a PyObject dict, you can do the same with C uint64s sitting on a much lower-level Cython view.

> Do you have benchmarks for these two variants?

Working on one now, hold on.

> Is this on performance critical path?

Yes, this is accessed for every chunk in present and past versions of a dataset when you exit the staged change context manager.

> * should there be any tests added for the `_commit_hash` implementations?

Impossible without writing test-specific Cython support modules and a test-specific meson file, which felt like overkill when this is already tested as part of #521.

crusaderky:
Benchmarks done:
```
| Change   | Before [1c10d295] <commit>   | After [189a29dd] <nocpp>   | Ratio   | Benchmark (Parameter)                                           |
|----------|------------------------------|----------------------------|---------|-----------------------------------------------------------------|
| +        | 1.95±0.01ms                  | 6.52±0.2ms                 | 3.35    | staged_changes.TimeCommitPlan.time_commit_plan('all_duplicate') |
| +        | 2.79±0.01ms                  | 11.5±0.5ms                 | 4.13    | staged_changes.TimeCommitPlan.time_commit_plan('all_new')       |
|          | 92.4±3ms                     | 92.8±2ms                   | 1.00    | staged_changes.TimeCommitPlan.time_commit_plan('fragmented')    |
| +        | 1.00±0.01ms                  | 3.72±0.06ms                | 3.70    | staged_changes.TimeCommitPlan.time_commit_plan('no_changes')    |
| +        | 3.30±0.04ms                  | 14.0±0.2ms                 | 4.23    | staged_changes.TimeCommitPlan.time_commit_plan('obsolete_base') |
| +        | 1.09±0.01ms                  | 3.76±0.06ms                | 3.45    | staged_changes.TimeCommitPlan.time_commit_plan('one_chunk')     |
```
`1c10d295` is this PR + #521 
`189a29dd` adds https://github.com/crusaderky/versioned-hdf5/commit/32a58c8ae21cc0fc5faf7ab57273b7a5ddea01fe on top
Benchmarks code at https://github.com/crusaderky/versioned-hdf5/commit/ff12263ed4237e306e154773052fd6db078e9169

All datasets are 10,200 chunks
obsolete_base has additional 3,600 chunks
| use case | map sets | map gets | notes |
| --- | --- | --- | --- |
| all_duplicate | 10,200 | 10,200 ||
| all_new | 20,400 | 10,200 ||
| fragmented | 10,200 | 0 | runtime is dominated by iteration over python lists |
| no_changes | 10,200 | 0 | missing a fast exit path here |
| obsolete_base | 23,700 | 10,200 ||
| one_chunk | 10,201 | 1 ||

versioned_hdf5/_commit_hash.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    bool operator==(const ChunkLoc& other) const noexcept {
+        return slab_idx == other.slab_idx && slab_offset == other.slab_offset;
+    }
+};
+
+struct ChunkLocHash {
+    std::size_t operator()(const ChunkLoc& key) const noexcept {
+        std::size_t h = std::hash<uint64_t>{}(key.slab_offset);
+        return h ^ (std::hash<uint64_t>{}(key.slab_idx) << 1u);
```
krassowski: This suggestion was generated with help from Opus 5. I drove it to build on top of your benchmarks to demonstrate where the hash collision can occur and then suggest a more robust alternative; that said, this is not blocking for this PR and could be considered separately in a follow-up.

```suggestion
        // This functor has to do all the mixing itself: libstdc++'s std::hash<uint64_t>
        // is the identity function, and std::unordered_map only reduces the result
        // modulo the bucket count.
        //
        // XOR-ing the two fields does not mix them. slab_offset is a multiple of
        // chunk_size[0] and slab_idx counts the slabs, so whenever the two span the
        // same bit range they cancel out. That is the normal case, not a corner case:
        // every __setitem__ stages its chunks onto a slab of its own, so a caller that
        // writes a large dataset one block at a time ends up with a roughly square
        // (n_slabs x chunks_per_slab) layout. On a 401x401 grid of chunks staged one
        // chunk row at a time, slab_offset ^ (slab_idx << 1) returns 1,984 distinct
        // values for 160,801 keys and every lookup walks ~118 nodes.
        //
        // Multiplying slab_idx by a large odd prime lifts it clear of the low bits,
        // which is all that is needed here. The prime is the one from FNV-1a
        // (Fowler/Noll/Vo), 2**40 + 2**8 + 0xb3; any large odd constant works.
        //
        // This is injective - it cannot collide at all - as long as
        //   slab_offset < PRIME              (1.1e12 rows), and
        //   slab_idx    < 2**64 / PRIME      (1.7e7 slabs)
        // so that neither term wraps into the other. Both bounds are orders of
        // magnitude above anything reachable in practice. If they ever stop holding,
        // wrap the result in a bit mixer with unconditional avalanche, e.g. David
        // Stafford's Mix13 (https://zimbry.blogspot.com/2011/09/better-bit-mixing-improving-on.html),
        // better known as the finaliser of Vigna's splitmix64, at roughly twice the cost.
        constexpr uint64_t PRIME = 0x100000001b3ULL;
        return static_cast<std::size_t>(key.slab_idx * PRIME + key.slab_offset);
```

krassowski (approved):
Thanks, your benchmark demonstrates that the cost of Python object creation is of high relevance here indeed. I would say let's merge it to unblock the rest of the work.

---

My question was on the edge of "is it worth it" because I am used to defaulting to Python's dict as a solution with a very good hash on average.

With help from Opus I re-run your benchmarks with addition of a few scenarios; maybe these are overblown edge-cases, I leave the judgment to you.

scenario | XOR (as-is on the branch) | Prime multiplication (see suggestion) | Python dict
-- | -- | -- | --
**blockwise_xl** | 1050 ms | 231 ms | 332 ms
blockwise | 5.5 ms | 3.8 ms | 12.9 ms
big_chunks (full commit()) | 96.4 ms | 107 ms | 107 ms

Details in https://gist.github.com/krassowski/a021a0831a90d401015d9fbd34c4c288.

versioned_hdf5/_commit_hash.hpp
crusaderky: Opus' verbosity sears my retinas once more. Nonetheless the change is correct.

PR #534, opened 2026-08-04
[SECURITY] Fix: delete_versions leaks data when writing corner chunks

crusaderky:
**Severity: LOW**

A user could
1. write sensitive data to version r1,
2. thoroughly overwrite the contents of the dataset on version r2,
3. call `delete_version(["r1"])`
4. compact database
5. ship the file to a third party that was not meant to see the contents of r1, or store it on low-security storage.

This PR fixes a bug where, when `delete_versions` overwrites a full chunk from r1 with a corner chunk (last chunk along 2+ axes) from r2 that doesn't align exactly with chunk_size, the partial contents of the chunk from r1 that exceed the valid surface of the corner chunk of r2 remain stored in raw_data.

### DIFF
--- a/tests/test_replay.py
+++ b/tests/test_replay.py
@@ -9,7 +9,6 @@ from hypothesis import given
 from hypothesis import strategies as st
 from ndindex import Slice
 from numpy.testing import assert_array_equal
-from versioned_hdf5.slicetools import spaceid_to_slice
 
 from versioned_hdf5 import VersionedHDF5File
 from versioned_hdf5.backend import DEFAULT_CHUNK_SIZE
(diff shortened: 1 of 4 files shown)


tests/test_replay.py
```diff
@@ -919,6 +920,67 @@ def test_recreate_virtual_dataset(vfile):
...
+    delete_versions(vfile, ["r0", "r1"])
+    expect_raw_data2 = [
+        [5, 0],  # r1 (1, 1) (corner)
+        [0, 0],
+        [6, 7],  # r2 (0, 0)
+        [8, 9],
+    ]
+
+    assert_array_equal(vfile["r2"]["values"][:], expect_vds)
+    assert_array_equal(vfile.f["_version_data/values/raw_data"], expect_raw_data2)
```
crusaderky, line 1023: before changes to replay.py, this line fails with

```
E        ACTUAL: array([[5, 0],
E              [0, 4],
E              [6, 7],
E              [8, 9]])
E        DESIRED: array([[5, 0],
E              [0, 0],
E              [6, 7],
E              [8, 9]])
```
Note point 4, which was leaked from r0.

tests/test_replay.py
```diff
@@ -9,6 +9,7 @@
 from hypothesis import strategies as st
 from ndindex import Slice
 from numpy.testing import assert_array_equal
+from versioned_hdf5.slicetools import spaceid_to_slice
```
krassowski: Did you intend to use it?

```suggestion
```

krassowski (approved):
For purpose of data leak fix this looks good to me.

It looks like cache invalidation does not cover some edge cases when multiple instances/constructs access the old versions on runtime. Below I'm attaching Opus generated tests cases that might be worth considering (either fixing or documenting as limitations one case-by-case basis) in the future.

<details>

```diff
diff --git a/tests/test_replay.py b/tests/test_replay.py
index 1cc4a03..6f9a3d2 100644
--- a/tests/test_replay.py
+++ b/tests/test_replay.py
@@ -920,10 +920,10 @@ def test_recreate_virtual_dataset(vfile):
     np.testing.assert_equal(orig_virtual_dataset, new_virtual_dataset)
 
 
-def test_delete_versions_invalidates_cache(vfile):
-    """delete_versions() shuffles the chunks of raw_data around. A version that was read
-    through the same VersionedHDF5File handle beforehand must not keep reading from the
-    stale chunk offsets afterwards.
+def setup_two_versions(vfile):
+    """Create two versions of a 3x3 dataset, the second of which alters a single chunk.
+
+    Return the expected contents of the second version.
     """
     with vfile.stage_version("r0") as sv:
         sv.create_dataset("values", data=np.arange(1, 10).reshape(3, 3), chunks=(2, 2))
@@ -932,6 +932,15 @@ def test_delete_versions_invalidates_cache(vfile):
 
     expect = np.arange(1, 10).reshape(3, 3)
     expect[0, 0] = 100
+    return expect
+
+
+def test_delete_versions_invalidates_cache(vfile):
+    """delete_versions() shuffles the chunks of raw_data around. A version that was read
+    through the same VersionedHDF5File handle beforehand must not keep reading from the
+    stale chunk offsets afterwards.
+    """
+    expect = setup_two_versions(vfile)
     assert_array_equal(vfile["r1"]["values"][:], expect)  # Populate the caches
 
     delete_versions(vfile, ["r0"])
@@ -939,6 +948,77 @@ def test_delete_versions_invalidates_cache(vfile):
     assert_array_equal(vfile["r1"]["values"][:], expect)
 
 
+@pytest.mark.xfail(
+    reason="VersionedHDF5File._version_cache is only invalidated when "
+    "delete_versions() is given the VersionedHDF5File instead of the bare h5py.File"
+)
+def test_delete_versions_invalidates_cache_raw_file(vfile):
+    """delete_versions() also accepts a bare h5py.File. The caches of a
+    VersionedHDF5File wrapping the same file must be invalidated all the same.
+
+    Note that only integer and timestamp keys are stored in
+    VersionedHDF5File._version_cache; string keys are not cached, so they are already
+    covered by clearing InMemoryGroup._instances.
+    """
+    expect = setup_two_versions(vfile)
+    assert_array_equal(vfile[0]["values"][:], expect)  # Populate _version_cache
+
+    delete_versions(vfile.f, ["r0"])
+
+    assert_array_equal(vfile[0]["values"][:], expect)
+
+
+@pytest.mark.xfail(
+    reason="only the VersionedHDF5File that delete_versions() was called on gets its "
+    "_version_cache invalidated"
+)
+def test_delete_versions_invalidates_other_handle(vfile):
+    """All VersionedHDF5File handles on a file must be invalidated by delete_versions(),
+    not just the one it was called on.
+    """
+    other = VersionedHDF5File(vfile.f)
+    expect = setup_two_versions(vfile)
+    assert_array_equal(other[0]["values"][:], expect)  # Populate _version_cache
+
+    delete_versions(vfile, ["r0"])
+
+    assert_array_equal(other[0]["values"][:], expect)
+
+
+@pytest.mark.xfail(
+    reason="clearing InMemoryGroup._instances only affects future lookups; objects "
+    "that were already handed out keep reading from the stale chunk offsets"
+)
+def test_delete_versions_invalidates_held_group(vfile):
+    """A version group obtained before delete_versions() must not keep reading from the
+    stale chunk offsets afterwards.
+    """
+    expect = setup_two_versions(vfile)
+    r1 = vfile["r1"]
+    assert_array_equal(r1["values"][:], expect)  # Populate the caches
+
+    delete_versions(vfile, ["r0"])
+
+    assert_array_equal(r1["values"][:], expect)
+
+
+@pytest.mark.xfail(
+    reason="clearing InMemoryGroup._instances only affects future lookups; objects "
+    "that were already handed out keep reading from the stale chunk offsets"
+)
+def test_delete_versions_invalidates_held_dataset(vfile):
+    """Same as test_delete_versions_invalidates_held_group, for a dataset that was
+    obtained directly.
+    """
+    expect = setup_two_versions(vfile)
+    values = vfile["r1"]["values"]
+    assert_array_equal(values[:], expect)  # Populate the caches
+
+    delete_versions(vfile, ["r0"])
+
+    assert_array_equal(values[:], expect)
+
+
 def test_delete_versions_resets_edge_chunk_padding(vfile):
     """raw_data must not retain any data from the deleted versions in the padding around
     the chunks that lie on the edge of the virtual dataset. This includes the corner of
```

</details>

crusaderky:
> VersionedHDF5File._version_cache is only invalidated when delete_versions() is given the VersionedHDF5File instead of the bare h5py.File

This functionality predates my arrival on the project. I think the purpose here is to avoid bothering a user that just wants to open the File, delete a version, and exit. For this reason I'm not terribly concerned about stale caches.

> All VersionedHDF5File handles on a file must be invalidated by delete_versions(), not just the one it was called on.

I think this is inconsequential from a real-life use standpoint

> clearing InMemoryGroup._instances only affects future lookups; objects that were already handed out keep reading from the stale chunk offsets

Nobody expects things to go well if you hold a reference to a version while you call delete_versions - and afterwards you pretend to work on that version.


I'll investigate the blast radius of using classvar weakrefs everywhere to solve the first two points.

crusaderky:
I've mostly fixed cache invalidation; only the use case of the user holding a reference to a Dataset remains broken

PR #543, opened 2026-08-12
Fix `@cachedproperty` deadlock on Python <3.12

crusaderky:
I did not add unit tests trying to replicate the issue. There are zero multi-threading unit tests today and they should be added organically.


OmkarAgnihotri:
Hi @crusaderky, thanks for the fix. Can we issue a new release with this fix ? This is currently blocking a couple of workflows which use versioned_hdf5 in a multithreaded environment.

CC'ing @ArvidJB.

crusaderky:
Yes, a new release is planned by end of this week

PR #552, opened 2026-08-27
Speed up fast get/set actions on StagedChangesArray

crusaderky:
Speed up use case where the user performs many small `__getitem__` or `__setitem__` calls on a InMemoryDataset or InMemorySparseDataset.

```
| Change   | Before [bc80e415] <master>   | After [67781086] <is_array_protocol>   | Ratio   | Benchmark (Parameter)                                          |
|----------|------------------------------|----------------------------------------|---------|----------------------------------------------------------------|
| -        | 493±60ms                     | 268±7ms                                | 0.54    | staged_changes.TimeGetItem.time_getitem('all_fragmented')      |
| -        | 560±100ms                    | 272±3ms                                | 0.49    | staged_changes.TimeSetItem.time_setitem('all_fragmented')      |
```

### DIFF
--- a/versioned_hdf5/tools.py
+++ b/versioned_hdf5/tools.py
@@ -6,7 +6,7 @@ import ndindex
 import numpy as np
 from numpy.typing import ArrayLike, DTypeLike
 
-from versioned_hdf5.typing_ import ArrayProtocol
+from versioned_hdf5.typing_ import is_array_protocol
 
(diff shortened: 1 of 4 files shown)


versioned_hdf5/tools.py
crusaderky, line 18: This function becomes very hot when the user quickly iterates on `StagedChangesArray.__getitem__` or `StagedChangesArray.__setitem__`

PR #547, opened 2026-08-12
Redesign the commit stage: switch InMemoryArrayDataset to the new commit backend

crusaderky:
Towards #398

Follow-up to #535, which switches InMemoryArrayDataset to use the new StagedChangesArray.commit(). 

### Out of scope
- `write_dataset` and `create_base_dataset` are now exclusively called with a dummy array of size 0. They can be drastically simplified. (in scope for phase 1)

These remain on the old slow path:
- `create_virtual_dataset` (in scope for phase 2)
- `delete_versions`, `update_metadata`, `VersionedHDF5.rebuild_hashtables` (in scope for phase 3)

### Benchmarks

The ASV benchmark shows major performance improvements compared to #535.
There are a handful of minor performance regressions, which I will address in follow-ups:

- `StagedChangesArray.from_array` unnecessarily deep-copies partial edge chunks in the event of a resize() later on, but the array is created immediately before commit so that will never happen. It can be optimized away.
 - `commit_staged_changes` unnecessarily creates `ndindex.Tuple` objects, which is expensive, in order to feed them to  `create_virtual_dataset`. This is a transitory artifact necessary for backwards compatibility, which will go away in phase 2.
- The hash of the full slab could be cached (this was already noted in #535)

```bash
pixi r asv-compare 5ded40d9fd7eae320211bd1a59c51a5dd71e74b2 60eeb055a548cd80c59945cb957eebead7689240
```

| Change   | Before [5ded40d9] <master~8>   | After [60eeb055] <arraydataset>   | Ratio   | Benchmark (Parameter)                                                                |
|----------|--------------------------------|-----------------------------------|---------|--------------------------------------------------------------------------------------|
|          | 753±50ms                       | 885±40ms                          | ~1.17   | many_chunks.time_many_chunks                                                         |
| -        | 1.48±0.09s                     | 834±20ms                          | 0.56    | many_chunks.time_many_chunks_arange                                                  |
|          | 776±70ms                       | 855±70ms                          | ~1.10   | many_chunks.time_many_chunks_integer_index                                           |
| +        | 34.3±2ms                       | 43.7±2ms                          | 1.27    | resize.time_resize                                                                   |
|          | 130±30ms                       | 115±30ms                          | ~0.88   | resize.time_resize_hdf5                                                              |
|          | 529±50ms                       | 639±30ms                          | ~1.21   | staged_changes.TimeCommit.time_commit('fragmented')                                  |
[...]

PR #554, opened 2026-09-01
StagedChangesArray.from_array: keep trimmed edges

crusaderky:
Towards https://github.com/deshaw/versioned-hdf5/issues/398

`StagedChangesArray.from_array(as_base_slabs=False)` used to eagerly deep-copy the edge chunks of the input NumPy array, to cater for the possibility that the array may be enlarged later on.
This PR makes the deep-copy lazy, making it happen inside `resize()` when and if it is actually invoked.

This speeds up the following use cases:

- a new dense dataset (`InMemoryArrayDataset`), with shape not exactly divisible by chunk size, is enlarged with `resize()`. This causes `DatasetWrapper` to hot-swap it for a `InMemorySparseDataset`. The dataset is not enlarged again before being committed.
- any dataset, with shape not exactly divisible by chunk size, is completely overwritten with `__setitem__`, which triggers a hot-swap to `InMemoryArrayDataset`, and then enlarged, which triggers a second hot-swap to `InMemorySparseDataset`; it is not enlarged again afterwards.
- _(the important one)_ downstream of #547, a new dense dataset, with shape not exactly divisible by chunk size, is committed.

# Benchmark

ASV shows how the deep-copy is pushed from StagedChangesArray.from_array to `resize()`:

| Change   | Before [43aa364a]   | After    | Ratio   | Benchmark (Parameter)                                               |
|----------|------------------------------|------------------------------------------------|---------|---------------------------------------------------------------------|
| -        | 943±200μs                    | 139±50μs                                       | 0.15    | staged_changes.TimeFromArray.time_from_array('staged')              |
| +        | 300±20μs                     | 2.94±0.2ms                                     | 9.82    | staged_changes.TimeResize.time_resize('append', 'from_array')       |
| +        | 640±100μs                    | 3.41±0.3ms                                     | 5.32    | staged_changes.TimeResize.time_resize('enlarge', 'from_array')      |

### DIFF
--- a/docs/staged_changes.rst
+++ b/docs/staged_changes.rst
@@ -324,6 +324,16 @@ be physically filled with the fill_value:
 2. then, there is a transfer from the full slab to the staged slab for the extra area
    that needs to be filled with the fill_value.
 
+Trimmed staged slabs
+^^^^^^^^^^^^^^^^^^^^
+``from_array(as_base_slabs=False)`` creates staged slabs as views of the input array.
+When the input array's shape is not exactly divisible by the ``chunk_size``, these views
+may not be large enough to contain the chunks after ``resize()`` enlarges the array.
+``resize()`` deep-copies trimmed slabs into new ones padded with empty data. It's
+important that this happens lazily, only when and if needed, to avoid unnecessary
+deep-copies when the array is never enlarged over its lifetime - notably, when it's
+created and then immediately consumed and destroyed when committing an
+InMemoryArrayDataset.
 
 ``load()`` algorithm
 --------------------
--- a/docs/staged_changes.rst
+++ b/docs/staged_changes.rst
@@ -333,7 +333,7 @@ may not be large enough to contain the chunks after ``resize()`` enlarges the ar
 important that this happens lazily, only when and if needed, to avoid unnecessary
 deep-copies when the array is never enlarged over its lifetime - notably, when it's
 created and then immediately consumed and destroyed when committing an
-InMemoryArrayDataset.
+``InMemoryArrayDataset``.
 
 ``load()`` algorithm
 --------------------
--- a/docs/staged_changes.rst
+++ b/docs/staged_changes.rst
@@ -324,6 +324,7 @@ be physically filled with the fill_value:
 2. then, there is a transfer from the full slab to the staged slab for the extra area
    that needs to be filled with the fill_value.
 
+
 Trimmed staged slabs
 ^^^^^^^^^^^^^^^^^^^^
 ``from_array(as_base_slabs=False)`` creates staged slabs as views of the input array.
--- a/tests/test_staged_changes.py
+++ b/tests/test_staged_changes.py
@@ -1364,8 +1364,7 @@ def test_from_array_as_staged_slabs_6():
 def test_from_array_as_staged_slabs_trimmed_shapes():
     """from_array(..., as_base_slabs=False)
 
-    Edge slabs are trimmed to the exact edge size, never padded,
-    and _has_trimmed_staged_slabs is set.
+    Edge slabs are trimmed to the exact edge size, never padded.
     """
     arr = np.arange(35).reshape((5, 7))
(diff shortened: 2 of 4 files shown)


docs/staged_changes.rst
```diff
@@ -324,6 +324,16 @@ be physically filled with the fill_value:
...
+Trimmed staged slabs
+^^^^^^^^^^^^^^^^^^^^
+``from_array(as_base_slabs=False)`` creates staged slabs as views of the input array.
+When the input array's shape is not exactly divisible by the ``chunk_size``, these views
+may not be large enough to contain the chunks after ``resize()`` enlarges the array.
+``resize()`` deep-copies trimmed slabs into new ones padded with empty data. It's
+important that this happens lazily, only when and if needed, to avoid unnecessary
+deep-copies when the array is never enlarged over its lifetime - notably, when it's
+created and then immediately consumed and destroyed when committing an
+InMemoryArrayDataset.
```
krassowski:
```suggestion
``InMemoryArrayDataset``.
```

tests/test_staged_changes.py
```diff
@@ -1303,6 +1361,151 @@ def test_from_array_as_staged_slabs_6():
     assert_array_equal(arr, [0, 1, 2])
 
 
+def test_from_array_as_staged_slabs_trimmed_shapes():
+    """from_array(..., as_base_slabs=False)
+
+    Edge slabs are trimmed to the exact edge size, never padded,
+    and _has_trimmed_staged_slabs is set.
```
krassowski: Leftover from an earlier iteration?

```suggestion
    Edge slabs are trimmed to the exact edge size, never padded.
```

docs/staged_changes.rst
```diff
@@ -324,6 +324,16 @@ be physically filled with the fill_value:
 2. then, there is a transfer from the full slab to the staged slab for the extra area
    that needs to be filled with the fill_value.
 
+Trimmed staged slabs
```
krassowski, line 328:
```suggestion

Trimmed staged slabs
```

krassowski (commented):
I see that the increase in the time is expected - it is mostly moving it from one place to another to enable the `InMemorySparseDataset` switch.

Image

However, it looks like this also has a memory cost for a narrow set of conditions (when expanding along axis 0 in a multiple not equal to that of a chunk size

```python
import tracemalloc

import h5py
import numpy as np

import versioned_hdf5

print("versioned_hdf5", versioned_hdf5.__version__)
data = np.zeros((10_000, 100))  # 7.63 MiB

with h5py.File("repro.h5", "w") as f:
    with VersionedHDF5File(f).stage_version("v1") as sv:
        ds = sv.create_dataset("x", data=data, chunks=(128, 100))
        tracemalloc.start()
        ds.resize((10_001, 100))
        peak = tracemalloc.get_traced_memory()[1]
        tracemalloc.stop()

print(f"data                    {data.nbytes / 2**20:6.2f} MiB")
print(f"allocated by ds.resize  {peak / 2**20:6.2f} MiB")
```

I ran this before/after:

```
# before
versioned_hdf5 2.4.0
data                      7.63 MiB
allocated by ds.resize    0.13 MiB
```

```
# after
versioned_hdf5 2.4.1.dev59+g60eeb055a
data                      7.63 MiB
allocated by ds.resize    7.74 MiB
```

My LLM suggests that this could be avoided.

Image

Is this something that should be addressed in this PR?

crusaderky:
> Is this something that should be addressed in this PR?

What you're suggesting is feasible, but requires _a lot_ of new logic which I can't honestly promise by end-of-week.
The use case it would cover is specifically when

a. a user creates a new dense dataset or completely updates in a single pass the whole contents of an existing one
b. aftewards, in this specific order, the user adds extra rows

it is a real use case, worth dealing with properly, but doesn't feel like extremely urgent to me?

krassowski (approved):
> it is a real use case, worth dealing with properly, but doesn't feel like extremely urgent to me?

Okay, let's track this in a new issue then.

PR #556, opened 2026-09-03
Reduce peak memory in recreate_dataset

crusaderky:
Fix performance regression introduced by #547, where `recreate_dataset` started loading each version of a dataset fully in memory.

| [3d4d08cc] (master)   | #547   |   This PR | Benchmark (Parameter)                                               |
|------------------------------|-----------------------------------|---------|---------------------------------------------------------------------|
| 151M                         | 151M                              |    151M | replay.Baseline.peakmem_baseline('baseline')                        |
| 151M                         | 169M                              |    151M | replay.TimeRecreateDataset.peakmem_recreate_dataset('drop_version') |
| 151M                         | 173M                              |    151M | replay.TimeRecreateDataset.peakmem_recreate_dataset('no_callback')  |
| 508±4ms                      | 195±7ms                           | 213±3ms | replay.TimeRecreateDataset.time_recreate_dataset('drop_version')    |
| 631±3ms                      | 257±4ms                           | 278±5ms | replay.TimeRecreateDataset.time_recreate_dataset('no_callback')     |

### DIFF
--- a/versioned_hdf5/backend.py
+++ b/versioned_hdf5/backend.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import datetime
+import itertools
 import logging
+import math
 import textwrap
 from collections.abc import Iterator
 from dataclasses import dataclass
@@ -14,16 +16,19 @@ from h5py._hl.selections import select
 from h5py._selector import Selector
 from ndindex import ChunkSize, Slice, Tuple
 
+from versioned_hdf5.cytools import ceil_a_over_b
 from versioned_hdf5.h5py_compat import HAS_NPYSTRINGS, h5py_astype
 from versioned_hdf5.hashtable import Hashtable
 from versioned_hdf5.slicetools import RawDataView
+from versioned_hdf5.staged_changes import StagedChangesArray
 from versioned_hdf5.typing_ import DEFAULT, Default
 
 if TYPE_CHECKING:
-    from versioned_hdf5.staged_changes import StagedChangesArray
     from versioned_hdf5.wrappers import FiltersMixin
 
 DEFAULT_CHUNK_SIZE = 2**12
+# Amount of RAM that rewrite_dataset() can allocate as a scratch area
+REWRITE_BUFFER_BYTES = 2**24  # 16 MiB
 DATA_VERSION = 4
 # data_version 2 has broken hashtables, always need to rebuild
 # data_version 3 hash collisions for string arrays which, when concatenated,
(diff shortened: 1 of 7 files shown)


crusaderky:
@krassowski rebased

versioned_hdf5/backend.py
```diff
@@ -14,16 +16,19 @@
...
+from versioned_hdf5.staged_changes import StagedChangesArray
 from versioned_hdf5.typing_ import DEFAULT, Default
 
 if TYPE_CHECKING:
-    from versioned_hdf5.staged_changes import StagedChangesArray
     from versioned_hdf5.wrappers import FiltersMixin
 
 DEFAULT_CHUNK_SIZE = 2**12
+# Amount of RAM that rewrite_dataset() can allocate as a scratch area
+REWRITE_BUFFER_BYTES = 2**24  # 16 MiB
```
krassowski: Why 16 MiB? On the toy datasets used in benchmarks it might seems like a fine choice, but going by the dataset sizes that were mentioned in other issues it looks too small (which means more overhead/slower rewrites). I would suggest
- increasing this to reduce the overhead for large datasets (64? 128? 256?) - if someone is using this library they are likely to have non-negligible GBs of RAM available (as the datasets themselves are large); for four versions of 512 MiB I saw a speedup of 25% (10 seconds) when switching to 256 MiB
- making this user-configurable (either by exposing `max_bytes` in public API of `recreate_dataset` and `modify_metadata`, maybe with a different name say `max_buffer_bytes`)
- some heuristic could be used for smaller datasets to keep it smaller when the performance gain is negligible, but I am not sure if it is worth extra complexity

If my assumption on the dataset sizes is incorrect, then just mentioning why it should be 16 MiB in the comment would be enough.

versioned_hdf5/backend.py
```diff
@@ -573,6 +578,115 @@ def empty(shape: tuple[int, ...], dtype) -> RawDataView:
...
+    into an unrelated `raw_data`, which may live in another file. This is what
+    `recreate_dataset` needs: it can't assume that the new hash table maps the chunks
+    to the same locations as the old one, even where the data is unchanged.
+
+    `data` is read one block of chunks at a time, so that peak memory usage is
+    O(max_bytes) instead of O(data.size). Deduplication is unaffected: each block is
+    deduplicated against the on-disk hash table, which by then already describes every
+    chunk written by the previous blocks and by the previous versions.
+
+    Parameters
```
krassowski, line 641: Nitpick `chunks` is missing in the list below.

versioned_hdf5/replay.py
krassowski, line 672: Pre-existing: this still loads everything into memory. `modify_metadata` does not seem to get much love in this package - and maybe there is a reason, if it is because it is rarely used, it probably should not block this PR; I opened https://github.com/deshaw/versioned-hdf5/issues/559 to track this.

versioned_hdf5/backend.py
krassowski: This is also relevant to https://github.com/deshaw/versioned-hdf5/issues/562.

krassowski (commented):
This looks good, much better than `master` wrt memory usage! Maybe it could be even better on three fronts:
- choosing a better buffer block size (tuning the tradeoff between memory usage and performance)
- expanding the scope to benefit the `modify_metadata` (this can be addressed separately, I opened https://github.com/deshaw/versioned-hdf5/issues/559)
- avoiding repeating some of the work which is now done on per-buffer block basis (https://github.com/deshaw/versioned-hdf5/issues/562)

I also opened https://github.com/deshaw/versioned-hdf5/issues/560 which was found while reviewing this PR - it highlights a regression in an edge case (compared to released `v2.4.0`, but not related to this PR directly), up to you whether that should block the release.

versioned_hdf5/backend.py
crusaderky: I'll increase to 64MB. Making it configurable feels overkill given how niche this functionality is.

versioned_hdf5/replay.py
crusaderky: modify_metadata is not a regression and definitely out of scope for this PR

PR #561, opened 2026-09-04
resize(): don't deep-copy a whole slab to enlarge just the last chunk

crusaderky:
#554 introduced a regression when a user calls `StagedChangedArray.from_array()` from a numpy array that is not exactly divisible by chunk size along axis 0, then calls `resize()` along axis 0 to enlarge. This caused the whole slab to be deep-copied just for the sake of giving the last chunk enough living space.

This PR fixes the regression by adding a special case for axis 0 to the machinery introduced by #554: skip the use case in `_untrim_staged_slabs`; then use ResizePlan to deep-copy just the offending last chunk to the new slab, and finally replace the slab with a truncated view of itself.

 | Benchmark | master | this PR |
 |---|---|---|
 | `staged_changes.TimeResize.time_resize('append_row', 'from_array', '8 kiB')` | 71.7ms | 3.38ms |
 | `staged_changes.TimeResize.time_resize('append_row', 'from_array', '512 kiB')` | 37.3ms | 2.26ms |
 | `staged_changes.TimeResize.time_resize('append_col', 'from_array', '8 kiB')` | 814μs | 819μs |
 | `staged_changes.TimeResize.time_resize('append_col', 'from_array', '512 kiB')` | 2.17ms | 2.07ms |
 | `staged_changes.TimeResize.time_resize('enlarge', 'from_array', '8 kiB')` | 73.2ms | 5.08ms |
 | `staged_changes.TimeResize.time_resize('enlarge', 'from_array', '512 kiB')` | 39.8ms | 5.77ms |
 | `staged_changes.TimeResize.time_resize('shrink', 'from_array', '8 kiB')` | 149μs | 161μs |
 | `staged_changes.TimeResize.time_resize('shrink', 'from_array', '512 kiB')` | 72.1μs | 75.5μs |
 | `staged_changes.TimeResize.time_resize_plan('append_row', 'from_array', '8 kiB')` | 1.06ms | 0.72ms |
 | `staged_changes.TimeResize.time_resize_plan('append_row', 'from_array', '512 kiB')` | 243μs | 390μs |
 | `staged_changes.TimeResize.time_resize_plan('append_col', 'from_array', '8 kiB')` | 182μs | 189μs |
 | `staged_changes.TimeResize.time_resize_plan('append_col', 'from_array', '512 kiB')` | 177μs | 177μs |
 | `staged_changes.TimeResize.time_resize_plan('enlarge', 'from_array', '8 kiB')` | 1.80ms | 1.86ms |
 | `staged_changes.TimeResize.time_resize_plan('enlarge', 'from_array', '512 kiB')` | 437μs | 588μs |
 | `staged_changes.TimeResize.time_resize_plan('shrink', 'from_array', '8 kiB')` | 143μs | 152μs |
 | `staged_changes.TimeResize.time_resize_plan('shrink', 'from_array', '512 kiB')` | 66.6μs | 72.9μs |

### DIFF
--- a/versioned_hdf5/staged_changes.py
+++ b/versioned_hdf5/staged_changes.py
@@ -479,6 +479,7 @@ class StagedChangesArray(MutableMapping[Any, T]):
             slab_offsets=self.slab_offsets.copy() if copy else self.slab_offsets,
             n_slabs=self.n_slabs,
             n_base_slabs=self.n_base_slabs,
+            slab_lengths=[0 if slab is None else slab.shape[0] for slab in self.slabs],
         )
 
     def _load_plan(self, copy: bool = True) -> LoadPlan:
@@ -676,6 +677,16 @@ class StagedChangesArray(MutableMapping[Any, T]):
 
             tplan.transfer(src_slab, dst_slab)
 
+        for slab_idx, new_size0 in plan.shrink_slabs:
+            assert slab_idx > self.n_base_slabs  # Never shrink base slabs
+            slab = self.slabs[slab_idx]
+            assert slab is not None
+            self.slabs[slab_idx] = slab[:new_size0]
+            ht = self.hash_tables[slab_idx]
+            if ht is not None:
+                new_nchunks = int(ceil_a_over_b(new_size0, self.chunk_size[0]))
+                self.hash_tables[slab_idx] = ht[:new_nchunks]
+
         for slab_idx in plan.drop_slabs:
             assert slab_idx != 0  # Never drop the full slab
             self.slabs[slab_idx] = None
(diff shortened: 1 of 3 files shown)


versioned_hdf5/staged_changes.py
```diff
@@ -479,6 +479,7 @@ def _resize_plan(self, shape: tuple[int, ...], copy: bool = True) -> ResizePlan:
             slab_offsets=self.slab_offsets.copy() if copy else self.slab_offsets,
             n_slabs=self.n_slabs,
             n_base_slabs=self.n_base_slabs,
+            slab_lengths=[0 if slab is None else slab.shape[0] for slab in self.slabs],
```
krassowski: This has a cost for all resizes (whether enlarging or shrinking). A way to avoid is to pass the slabs directly and only create the list when needed (in `_relocate_trimmed_staged_chunks_along_axis0`).

versioned_hdf5/staged_changes.py
```diff
@@ -1915,6 +1973,138 @@ def _enlarge_along_axis(
...
+
+        The relocations are one TransferPlan per source slab, defined by directly
+        invoking the plain constructor with read_many_slices() parameters computed
+        in bulk, bypassing the expensive from_mappers(); each trimmed slab typically
+        contributes a single tiny chunk, so the per-chunk index arithmetic of the
+        mappers would cost more than the copy itself.
+
+        When also enlarging along axis 1+, slabs that are also trimmed along that axis
+        don't reach this method: they were already deep-copied and padded by
+        _untrim_staged_slabs before the plan was created.
```
krassowski, line 2015: `self._untrim_staged_slabs(shape)` is called after creating the plan with `plan = self._resize_plan(shape, copy=False)`. Does this docstring require updating?

krassowski (approved):

versioned_hdf5/staged_changes.py
krassowski: I asked Opus 5 to benchmark how much of a difference this makes and it looks like this can have a noticeable impact for a loop with multiple appends and it is quadratic cost so becomes visible as extra seconds at around 8000 appends on my machine. The existing benchmarks do not cover it very well.

versioned_hdf5/staged_changes.py
crusaderky: That's not right. It's unwittingly deep-copying last chunks that have already been enlarged along with the rest of the slab - so deep-copying them twice. Good catch!

PR #567, opened 2026-09-16
Redesign the commit stage: rewrite `create_virtual_dataset` in Cython

crusaderky:
This PR delivers #398 phase 2.

Rewrite `create_virtual_dataset` in Cython. Instead of creating and then consuming the legacy `data_dict` of Tuple of Slice objects, which is very expensive to create and parse, directly scan the `slab_indices` and `slab_offsets` arrays in a tight C-only loop.

## Follow-up

These functions are no longer used by anything and can be removed. I deliberately left them untouched to keep the size of the PR contained; they'll be cleaned up in a follow-up PR.

- `backend.create_virtual_dataset`
- `staged_changes.StagedChanges.changes`
- `staged_changes.ChangesPlan`
- `wrappers.InMemoryDataset.data_dict`
- `wrappers.InMemorySparseDataset.data_dict`
- `wrappers._staged_changes_to_data_dict`

## Benchmarks

| Change   | v2.5.0                       | This PR                     | Ratio   | Benchmark (Parameter)                                                                                             |
|----------|------------------------------|-----------------------------|---------|-------------------------------------------------------------------------------------------------------------------|
| -        | 92.6±4ms                     | 27.1±4ms                    | 0.29    | chunk_guessing.TimeSetitemVsCreateDataset.time_one_element_commit('create_dataset')                               |
| -        | 40.9±2ms                     | 12.0±2ms                    | 0.29    | chunk_guessing.TimeSetitemVsCreateDataset.time_one_element_commit('setitem')                                      |
| -        | 946±40ms                     | 383±40ms                    | 0.40    | many_chunks.time_many_chunks                                                                                      |
|          | 1.42±0.1s                    | 1.17±0.05s                  | ~0.82   | many_chunks.time_many_chunks_arange                                                                               |
| -        | 916±60ms                     | 350±60ms                    | 0.38    | many_chunks.time_many_chunks_integer_index                                                                        |
| -        | 35.9±3ms                     | 25.7±2ms                    | 0.72    | resize.time_resize                                                                                                |
| -        | 1.90±0.01s                   | 957±30ms                    | 0.50    | resize.time_resize_and_write                                                                                      |
| -        | 92.5±7ms                     | 42.2±6ms                    | 0.46    | wrappers.TimeCommit.time_commit('v1_dense', (25, 25))                                                             |
| -        | 34.1±3ms                     | 20.6±2ms                    | 0.60    | wrappers.TimeCommit.time_commit('v1_dense', (50, 50))                                                             |
[...]

PR #571, opened 2026-09-21
Redesign the commit stage: phase 2 cleanup

crusaderky:
Follow-up to #567. Remove dead code.


tests/test_backend.py
```diff
@@ -371,201 +370,6 @@ def test_write_dataset_offset_multidimension(h5file):
     assert ds.dtype == np.float64
 
 
-@pytest.mark.setup_args(version_name="test_version")
```
crusaderky, line 374: All these tests were duplicated in #567

crusaderky:
FYI @krassowski

PR #572, opened 2026-09-21
rewrite_dataset() to reuse the in-memory hash table

crusaderky:
This PR is a follow-up to #567, fixing an issue where `rewrite_dataset` writes the partial hash table to disk, then reads it back up at every block; at each iteration, the amount of data read back increases.

This turns out to be a purely hypothetical worry - it is completely invisible in unit tests, even with tiny 1kiB chunks.
This PR is here because
- it was already finished by the time I realised it had no practical benefit;
- it's not _wrong_ and doesn't have any negative impact;
- most importantly, it will hopefully save an engineer time in the future from chasing this same red herring.

# Benchmarks
Zero impact across all pre-existing benchmarks.
A new benchmark that stresses this thing showed a microscopic and noisy performance increase, but only for gigantic datasets made up of minuscule chunks. It can be seen in 4745d93b83ee630a09043373cab5d0620120dd7c.

The only way to meaningfully show the performance benefit of this change is to monkey-patch the hardcoded 64 MiB block size to a handful of kilobytes, which is not representative of any real-life use case.

# Related

https://github.com/deshaw/versioned-hdf5/pull/567#discussion_r4045944306


crusaderky:
Superseded by #573

PR #573, opened 2026-09-23
`rewrite_dataset()`: stream blocks as views

crusaderky:
- Follow-up to #567
- Supersedes #572

Implements the suggestion in https://github.com/deshaw/versioned-hdf5/pull/567#discussion_r4045944306: `rewrite_dataset()` now commits one block of `data` at a time through a throwaway `StagedChangesArray.from_array(..., as_base_slabs=False)`, then copies the block's chunk map into the full-sized map.

This is faster than staging into the full-sized array:

- no copy: the block read from `data` as a NumPy array *is* the staged slabs, as views;
- per-block plans now see the block's chunk map instead of the whole array's, so the plan construction is O(block) instead of O(total) per block.

## Benchmarks

| Benchmark (Parameter) | master | This PR | Ratio |
|---|---|---|---|
| `replay.TimeRecreateDataset.time_recreate_dataset('drop_version')` | 376±2ms | 331±4ms | 0.88 |
| `replay.TimeRecreateDataset.time_recreate_dataset('no_callback')` | 514±6ms | 461±9ms | 0.90 |


krassowski (approved):
LGTM!

PR #574, opened 2026-09-23
Fix AssertionError in DatasetWrapper.resize() after sv[name] = arr

### LINKED ISSUE #569: `DatasetWrapper.resize()` raises `AssertionError` when enlarging a dataset that `sv[name] = arr` replaced in the same version
Found while fuzzing the public API for the review of #567. It has been there since #446 introduced the conversion to a sparse dataset on enlarge (v2.1.0); before the assertion was added the same script failed with a `TypeError` from the `StagedChangesArray` constructor.

```python
import h5py, numpy as np

with h5py.File("bug8.h5", "w") as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version("v0") as sv:
        sv.create_dataset("x", data=np.arange(10.0), chunks=(4,))

with h5py.File("bug8.h5", "r+") as f:
    vf = VersionedHDF5File(f)
    with vf.stage_version("v1") as sv:
        sv["x"] = np.arange(10.0) + 1  # replaces the whole dataset
        sv["x"].resize((12,))
```

Raises

```python
  File "bug8.py", line 13, in <module>
    sv["x"].resize((12,))
    ~~~~~~~~~~~~~~^^^^^^^
  File "versioned_hdf5/wrappers.py", line 1251, in resize
    new_ds = InMemorySparseDataset(
        name=self.dataset.name,
    ...<5 lines>...
        attrs=self.dataset.attrs,
    )
  File "versioned_hdf5/wrappers.py", line 1075, in __init__
    assert isinstance(chunks, tuple)
           ~~~~~~~~~~^^^^^^^^^^^^^^^
AssertionError
```

Tested with:
- #567 at d407346
- v2.5.0 from conda-forge
- v2.4.0 and v2.3.1 from conda-forge: `AssertionError`
- v2.1.0 from conda-forge: `TypeError: Argument 'chunk_size' has incorrect type (expected tuple, got NoneType)` at `wrappers.py:1210`

crusaderky:
Fixes #569.

`DatasetWrapper.resize()` raised `AssertionError` when enlarging a dataset that `sv[name] = arr` had replaced wholesale in the same version, because the replacement `InMemoryArrayDataset` had no chunk size.

The same use case also dropped the replaced dataset's `fillvalue` and user `attrs`; a non-default fillvalue made the commit fail and attributes were silently lost.

Multi-dimensional replacement still requires a chunk size to be known (from a previous version, or via `create_dataset(chunks=...)`): a multi-dimensional `sv[name] = arr` has no way to guess one and fails with the same `NotImplementedError` as committing it does.

All new tests fail before the fix.


versioned_hdf5/wrappers.py
```diff
@@ -1220,12 +1273,29 @@ def resize(self, size, axis=None):
...
+            raw_data = self.dataset.parent._pinned_raw_data(self.dataset.name)
+            if raw_data is not None:
+                pinned = tuple(raw_data.attrs["chunks"])
+                if len(pinned) == self.dataset.ndim:
+                    chunks = pinned
+        if chunks is None:
+            # No chunk size is pinned (e.g. a brand new ``group[name] = array``
+            # dataset); guess one like commit_version() would.
+            chunks = normalize_chunks(
+                None, self.dataset.shape, self.dataset._buffer.dtype
```
krassowski: It looks like using the new shape could be more performant?

```suggestion
                None, new_shape, self.dataset._buffer.dtype
```

versioned_hdf5/wrappers.py
```diff
@@ -446,6 +476,25 @@ def _recursion_to_root(
...
+        # "versions": a dataset may itself live in a subgroup with that name.
+        version_group = self.versioned_root
+        version_data = version_group.parent.parent
+        rel = posixpath.relpath(
+            posixpath.join(self.name, posixpath.basename(name)), version_group.name
+        )
+        try:
+            return version_data[posixpath.join(rel, "raw_data")]
+        except KeyError:
+            return None
```
krassowski: Related but pre-existing: https://github.com/deshaw/versioned-hdf5/issues/578

